PR #161, opened 2018-12-27
PBM-150: Fix 'TestListAgents' failures

timvaillancourt:


cli/pbmctl/main_test.go
```diff
@@ -61,72 +62,80 @@ func TestListAgents(t *testing.T) {
...
 		}
-		if client.Id != want[i].Id {
+		if client.NodeType != "NODE_TYPE_MONGOS" && client.Id != want[i].Id {
 			t.Errorf("Invalid client id. Got %v, want %v", client.Id, want[i].Id)
+		} else if client.NodeType == "NODE_TYPE_MONGOS" {
+			// fix for mongos having hostname:port instead of ip:port
+			hostname, _ := os.Hostname()
+			wantHostPort := strings.SplitN(want[i].Id, ":", 2)
+			wantHostPortId := hostname + ":" + wantHostPort[1]
+			if client.Id != wantHostPortId && client.Id != want[i].Id {
```
timvaillancourt: @percona-csalguero: I wanted to check this fix is valid or if the Id is wrong.

Mongos is returning "<hostname>:10000" on my test environment when the test expects "<ip>:10000".

PR #237, opened 2019-06-05
Add extra checks and avoiding panic

nonemax:


internal/oplog/oplog_apply.go
```diff
@@ -288,9 +288,28 @@ func processCreateIndex(sess *mgo.Session, buf []byte) error {
...
+		if int64Key, ok := key.Value.(int64); ok && int64Key < 0 {
+			sign = "-"
+		}
+		if float32Key, ok := key.Value.(float32); ok && float32Key < 0 {
+			sign = "-"
+		}
+		if floatKey, ok := key.Value.(float64); ok && floatKey < 0 {
+			sign = "-"
+		}
+
```
dAdAbird: As for me, type switch looks more clear and definitely more idiomatic:
```go
switch k := key.Value.(type) {
case int: 
    if k < 0 {
        sign = "-"
    }
case int8: 
    if k < 0 {
        sign = "-"
    }
case int16: 
    if k < 0 {
        sign = "-"
    }
}
...

```

PR #244, opened 2019-06-09
PBM-218 Filter for oplog apply

percona-csalguero:
New filter to ignore system collections when re-applying the oplog

### DIFF
--- a/internal/oplog/oplog_apply.go
+++ b/internal/oplog/oplog_apply.go
@@ -170,8 +170,12 @@ func (oa *Apply) Run() error {
 		}
 
 		if ns, ok := dest["ns"]; ok {
-			if skip(ns.(string)) {
-				continue
+			if nss, ok := ns.(string); ok {
+				if skip(nss) {
+					continue
+				}
+			} else {
+				return fmt.Errorf("invalid type for oplog cmd ns. Want string, got %T", ns)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


internal/oplog/oplog_apply.go
```diff
@@ -151,15 +152,29 @@ func (oa *Apply) Run() error {
...
 		}
 
+		if atomic.LoadUint32(&oa.debug) != 0 {
+			fmt.Printf("%#v\n", dest)
+		}
+
+		if ns, ok := dest["ns"]; ok {
+			if skip(ns.(string)) {
+				continue
+			}
```
dAdAbird: I believe using the map would be a cleaner and simpler solution

```go
// global scope
var systemNS = map[string]struct{}{
	"config.system.sessions":   struct{}{},
	"config.cache.collections": struct{}{},
}

...
// func (oa *Apply) Run
if ns, ok := dest["ns"]; ok {
	if _, ok := systemNS[ns.(string)]; ok {
		continue
	}
}
```

internal/oplog/oplog_apply.go
```diff
@@ -151,15 +152,29 @@ func (oa *Apply) Run() error {
...
 		if !ok {
 			return fmt.Errorf("invalid oplog document. there is no command")
 		}
 
+		if atomic.LoadUint32(&oa.debug) != 0 {
+			fmt.Printf("%#v\n", dest)
+		}
+
+		if ns, ok := dest["ns"]; ok {
+			if skip(ns.(string)) {
```
dAdAbird: Just in case, are we sure `ns` is always a `string`?

percona-csalguero: yes, it is the database name

PR #255, opened 2019-06-12
PBM-196 Update list-nodes and list-storages

nonemax:


grpc/server/server.go
```diff
@@ -316,39 +312,15 @@ func (s *MessagesServer) listStorages() (map[string]StorageEntry, error) {
 	var errs error
 	l := &sync.Mutex{}
```
dAdAbird: Looks like there is no more need in this variable

PR #257, opened 2019-06-13
Pbm-178 and pbm-246

nonemax:


tplavcic:
I'm not sure if change in the logger could brake how docker images and packaging works, maybe better leave this until after this release.

PR #258, opened 2019-06-13
PBM-248 Enable SSL support for mongo sessions

percona-csalguero:
Also added a new channel to signal the agent's stop to the db watchdog.
Cannot use the context.Done() because the context is being re-initialized upon reconnection.

### DIFF
--- a/grpc/client/client.go
+++ b/grpc/client/client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/globalsign/mgo"
 	"github.com/globalsign/mgo/bson"
+	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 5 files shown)


grpc/client/client.go
```diff
@@ -527,6 +530,10 @@ func (c *Client) processIncommingServerMessages() {
 func (c *Client) dbWatchdog() {
 	for {
 		select {
+		case <-c.watchDogStopChan:
+			return
```
dAdAbird: This will prevent re-registering the client after it being stopped.
The watchdog should always work while the agent is running. Otherwise, fix for https://jira.percona.com/browse/PBM-239 (https://github.com/percona/percona-backup-mongodb/commit/8114b8f8471ff19cba6acbb23bd7c8e8ea649c65) won't work.

PR #282, opened 2019-10-13
Changed read/write layout 

dAdAbird:
In a simplicity sake and better error handling

### DIFF
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -60,6 +60,8 @@ func (b *Backup) Run(bcp pbm.BackupCmd) (err error) {
 		if err != nil {
 			rsMeta.Status = pbm.StatusError
 			rsMeta.Error = err.Error()
+			meta.Status = rsMeta.Status
+			meta.Error = rsMeta.Error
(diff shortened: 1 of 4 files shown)


pbm/backup/backup.go
```diff
@@ -122,16 +128,98 @@ func (b *Backup) Run(bcp pbm.BackupCmd) (err error) {
...
+		return false, errors.Wrap(err, "get node replication lag")
+	}
+
+	return replLag < maxReplicationLagTimeSec && status.Health == pbm.NodeHealthUp &&
+			(status.State == pbm.NodeStatePrimary || status.State == pbm.NodeStateSecondary),
+		nil
+}
+
+// RWErr multierror for the read/compress/write-to-store operations set
+type RWErr struct {
```
delgod: Do we need to expose this error?
If yes may we use one of the options:
- Expose pointer instead of value. This is more useful in errors.Cause(err).(struct) comparison.  
- Create IsRWErr func where we can hide logic

dAdAbird: Actually, yes, no need to make this error type exported now

PR #288, opened 2019-10-24
PBM-361 - use logical clock for oplog tailing

dAdAbird:
also add more error handling for the oplog tailing


dAdAbird:
Despite using the primitive bson change nothing in logic it makes code more certain and clear.

PR #281, opened 2019-10-09
Add Amazon S3 backup folder to bucket.

drafil:
Added Amazon S3 backup folder to filesystem path. This allows a better backup separation by folder inside the same bucket.

PR #289, opened 2019-10-24
Added optional backup destination name.

drafil:
The new optional backup_name argument allows users to specify the destination backup name relative to the S3 bucket. It can be a file name or a file path, in which case the corresponding directory structure will be created in the bucket. If the argument is not used, the file name defaults to the current timestamp and it is uploaded to the bucket root (the original behaviour).


dAdAbird:
Hi,

I'd like to stay with the design described here https://github.com/percona/percona-backup-mongodb/pull/281#pullrequestreview-300115685. I still don't think that having a path in the backup name is a good idea.

drafil:
@dAdAbird 

Hi Andrew,

The problem with the requirements for #281 is that AWS S3 bucket names are unique DNS-compliant names and they cannot be expressed as paths. I believe it would be misleading for the user to define a bucket in the config as a directory structure:
https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html
create_bucket
invalid_bucket_name

On the other hand, I think it is very useful to be able to logically separate backups in the same bucket by environment and cluster name (this is my use case).

Are #281 design requirements hard for a specific reason? If no, would it be a better solution to add an extra config field for the destination directory? E.g.:
```yaml
type: s3
s3:
    region: us-east-1
    bucket: mongo-backups
    directory: dev/my-sharded-cluster
```

dAdAbird:
HI @drafil,

Adding the extra config field looks good. Just name it `prefix` instead of `folder.

I reopen the PR, but you might want to merge the upstream before further work since some changes regarding the config have been committed recently. Or you may close this PR and open the new one.

drafil:
I will close this PR and open a new one since many modifs took place in the meantime.

PR #292, opened 2019-10-29
PBM-363

dAdAbird:


pbm/backup/oplog.go
```diff
@@ -23,24 +24,45 @@ func NewOplog(node *pbm.Node) *Oplog {
...
-	cur, err := cl.Find(ctx, bson.M{
-		"op": bson.M{"$ne": pbm.OperationNoop},
-		"ts": bson.M{"$gte": fromTs},
-	})
+	cur, err := cl.Find(ctx,
+		bson.M{
+			"ts": bson.M{"$gte": from},
+		},
+		options.Find().SetCursorType(options.Tailable),
+	)
```
akira-kurogane: Shouldn't `"ts": bson.M{"$lte": to}` be in the query args here as well?
This would make it possible to remove the `primitive.CompareTimestamp(to, opts) > 0` check in the cursor iteration below too.

pbm/backup/oplog.go
```diff
@@ -23,24 +24,45 @@ func NewOplog(node *pbm.Node) *Oplog {
...
+	opts := primitive.Timestamp{}
+	var ok bool
 	for cur.Next(ctx) {
+		opts.T, opts.I, ok = cur.Current.Lookup("ts").TimestampOK()
+		if !ok {
+			return errors.Errorf("get the timestamp of record %v", cur.Current)
+		}
+		if primitive.CompareTimestamp(to, opts) > 0 {
+			return nil
+		}
```
akira-kurogane: (See comment on lines 39-44)

pbm/backup/oplog.go
akira-kurogane: Ah, I think that must be because the tailable cursor option is still on. I didn't notice it yesterday, so sorry.

As we're starting the query after all the target records are present, already, we don't have to tail any more. Just dump in one pass.

dAdAbird: I turned cursor intentionally. Because I'm not sure that we're starting the query after all the target records are present. We just start the query after the latest dump was finished. And there are no guarantees that no targeted records gonna came after we started the tailing (due network issues, node lag etc.). Ok, the probability of such behaviour not so high. But I'd prefer to be secure in that matter.

PR #298, opened 2019-11-08
PBM-344 Store -> config

dAdAbird:


pbm/config.go
```diff
@@ -0,0 +1,176 @@
...
+	return false
+}
+
+func (p *PBM) GetConfigYaml(safe bool) ([]byte, error) {
+	c, err := p.GetConfig()
+	if err != nil {
+		errors.Wrap(err, "get from db")
+	}
+
+	if safe {
```
akira-kurogane: It would be clearer to call "safe" "fieldRedaction"

PR #302, opened 2019-11-24
PBM-356 deadlocks prevention

dAdAbird:


agent/agent.go
```diff
@@ -112,13 +117,14 @@ func (a *Agent) Backup(bcp pbm.BackupCmd) {
 	err = backup.New(a.pbm, a.node).Run(bcp)
 	if err != nil {
 		log.Println("[ERROR] backup:", err)
+	} else {
+		log.Printf("Backup %s finished", bcp.Name)
```
akira-kurogane: Shouldn't the log message for the user be 'backup xxx started' rather than 'backup xxx finished'?

dAdAbird: "Backup started" message is few lines above. At this, point backup is finished (if `Run(bcp)` returned no error).

akira-kurogane: Ah, I think I see. This would only be in the very untypical case the whole backup completes in < StaleFrameSec.

PR #303, opened 2019-11-25
PBM-362 Work with standalone node

dAdAbird:
Make possible to backup/restore from standalone node


dAdAbird:
see https://github.com/percona/percona-backup-mongodb/pull/305 and comments for the ticket

PR #305, opened 2019-11-26
No standalone backups but the single-node replset instead

dAdAbird:


pbm/backup/backup.go
```diff
@@ -39,7 +39,7 @@ func (b *Backup) Run(bcp pbm.BackupCmd) (err error) {
 	return b.run(bcp)
 }
 
-var statusRunTimeout = time.Second * 15
+var wiatBackupStart = time.Second * 15
```
delgod:
```suggestion
var waitBackupStart = time.Second * 15
```

PR #306, opened 2019-11-27
PBM-369 Preserve parameters from user provided URL

dAdAbird:
Except for the `replicaSet` which causes an error while reconnecting to the ConfigServer on a sharded cluster


akira-kurogane:
OK, reworked. Especially as I realized a replicaset connection was needed by the "pbm" class for non-sharded rs nodes and cfgsvr rs nodes too. Because the pbm class is the one with ListenCmd() function.

PR #311, opened 2019-12-05
Fix locking for replicasets and fast backups

dAdAbird:


agent/agent.go
```diff
@@ -114,16 +114,29 @@ func (a *Agent) Backup(bcp pbm.BackupCmd) {
...
+	// In the case of fast backup (small db) we have to wait before releasing the lock.
+	// Otherwise, since the primary node waits for `WaitBackupStart*0.9` before trying to acquire the lock
+	// it might happen that the backup will be made twice:
+	//
+	// secondary1 >---------*!lock(fail - acuired by s1)---------------------------
+	// secondary2 >------*lock====backup====*unlock--------------------------------
+	// primary    >--------*wait--------------------*lock====backup====*unlock-----
+	//
+	// Secondaries also may start trying to acquire a lock with quite an interval (e.g. due to network issues)
+	// TODO: we cannot relay on the nodes wall clock.
```
akira-kurogane:
```suggestion
	// TODO: we cannot rely on the nodes wall clock.
```
Just a spelling change

agent/agent.go
```diff
@@ -114,16 +114,29 @@ func (a *Agent) Backup(bcp pbm.BackupCmd) {
...
+	// TODO: ? pbmBackups should have unique index by name ?
```
akira-kurogane: Seems like a good idea. Would adding the index also prevent this double-backup possibility?

dAdAbird: yep, this is my intention. I made a ticket in JIRA

PR #314, opened 2019-12-18
PBM-385 Resync backups meta

dAdAbird:


cmd/pbm/main.go
```diff
@@ -20,11 +20,12 @@ var (
 	pbmCmd = kingpin.New("pbm", "Percona Backup for MongoDB")
 	mURL   = pbmCmd.Flag("mongodb-uri", "MongoDB connection string").String()
 
-	configCmd     = pbmCmd.Command("config", "Set, change or list the config")
-	configListF   = configCmd.Flag("list", "List current settings").Bool()
-	configFileF   = configCmd.Flag("file", "Upload config from YAML file").String()
-	configSetF    = configCmd.Flag("set", "Set the option value <key.name=value>").StringMap()
-	configShowKey = configCmd.Arg("key", "Show the value of a specified key").String()
+	configCmd           = pbmCmd.Command("config", "Set, change or list the config")
+	configRsyncBcpListF = configCmd.Flag("resync", "Resync backup list with the current store").Bool()
```
akira-kurogane: Just an idea (and not a blocker): maybe this should be "force-resync" so users don't get the idea that it is something they're obliged to do in normal circumstances.

pbm/rsync.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	_, err = p.Conn.Database(DB).Collection(BcpCollection).InsertMany(p.ctx, ins)
+	if err != nil {
+		return errors.Wrap(err, "insert retrieved backups meta")
+	}
+
+	return nil
+}
+
+func (p *PBM) copyBackupsMeta(bm []BackupMeta) error {
```
akira-kurogane: Would "archiveBackupsMeta" be an OK function name?

PR #317, opened 2020-01-07
Fixes for PBM-391, PBM-392 and PBM-394

dAdAbird:


cmd/pbm/restore.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Stop if there is some live operation.
+	// But if there is some stale lock leave it for agents to deal with.
+	for _, l := range locks {
+		if l.Heartbeat.T+pbm.StaleFrameSec >= ts.T {
+			return errors.Errorf("another operation in progress, %s/%s", l.Type, l.BackupName)
+		}
+	}
+
+	bcp, err := cn.GetBackupMeta(bcpName)
```
delgod: I think it is better to check backup first and lock exactly before SendCmd - in such case it is harder to get race condition from different pbmctl

dAdAbird: What exactly scenario do you mean by the race condition in from a different pbmctl?

delgod: I mean when we run restore command in parallel in two terminals exactly in the same second

PR #325, opened 2020-01-28
v1.1.1 fixes + Auto tests

dAdAbird:


dAdAbird:
> Something for later though is I noticed that mgo driver vendor pkg is still included. Is there an entanglement in another vendor package to mgo?

Oh, nice catch. This is an autocomplete has imported wrong package in the tests. I'll fix it later on.

PR #326, opened 2020-01-31
PBM-414 Use minio-client for GCS

dAdAbird:


go.mod
```diff
@@ -7,16 +7,19 @@ require (
 	github.com/alecthomas/kingpin v2.2.6+incompatible
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
-	github.com/aws/aws-sdk-go v1.25.19
+	github.com/aws/aws-sdk-go v1.28.9
```
delgod: is it still needed???

pbm/backup/dst.go
```diff
@@ -48,30 +49,41 @@ func Save(data io.Reader, stg pbm.Storage, name string) error {
...
+				u.LeavePartsOnError = true    // Don't delete the parts if the upload fails.
+				u.Concurrency = 1
+			}).Upload(&s3manager.UploadInput{
+				Bucket: aws.String(stg.S3.Bucket),
+				Key:    aws.String(path.Join(stg.S3.Prefix, name)),
+				Body:   data,
+			})
+			return errors.Wrap(err, "upload to S3")
+		case pbm.S3ProviderGCS:
+			// using minio client with GCS because it
```
delgod: we can use minio client always

pbm/config.go
```diff
@@ -82,13 +92,18 @@ func (p *PBM) SetConfigByte(buf []byte) error {
...
+	_, err = p.Conn.Database(DB).Collection(ConfigCollection).UpdateOne(
 		p.ctx,
 		bson.D{},
 		bson.M{"$set": cfg},
-		options.Update().SetUpsert(true))
-
-	return err
+		options.Update().SetUpsert(true),
+	)
+	return errors.Wrap(err, "mongo UpdateOne")
```
akira-kurogane: How about this for providing some context that will make the error string more unique?, if it shows up in JIRA tickets etc.
```suggestion
	return errors.Wrap(err, "mongo ConfigCollection UpdateOne")
```

pbm/backup/dst.go
dAdAbird: let's do it in the next release

PR #327, opened 2020-01-31
PBM-416: Release notes - 1.1.1

borys-belinsky:


doc/source/release-notes/1.1.1.rst
```diff
@@ -0,0 +1,38 @@
+.. _PBM-1.1.0:
```
akira-kurogane: I noticed this page structure header is 1.1.0 not 1.1.1.

If I recall the last release you told me this is something that gets fixed as part of the documentation release process automatically. If that is the case please ignore.

PR #339, opened 2020-03-03
Adding doc note explaining new collections not dropped during restore

akira-kurogane:


akira-kurogane:
Hi @borys-belinsky. This is documentation patch is for PBM-428.

In a sense it is PBM-234 being done again. I inadvertently removed that ticket's doc note during the v1.1 documentation rewrite it seems.

PR #346, opened 2020-03-17
it's no need to wait because all agents are not started at the same t…

leoloel:
…ime, and the processing speed of each machine is different


leoloel:
> @leoloel could you please don't change tabs to spaces?

 change tabs to spaces? I don't know what you mean.

delgod:
> > @leoloel could you please don't change tabs to spaces?
> 
> change tabs to spaces? I don't know what you mean.

sorry, it is not spaces issue, it is unix(LF)<->dos(CR-LF) issue
please revert utf-8[dos] encoding back

PR #377, opened 2020-05-13
PBM-474 Added Ubuntu 20.04 to the supported platforms list

nastena1606:


doc/source/installation.rst
```diff
@@ -65,8 +65,8 @@ Installing |pbm| Using ``apt``
 * Debian 9 ("stretch")
 * Ubuntu 14.04 LTS (Trusty Tahr)
```
EvgeniyPatlan, line 66: it is not supported

doc/source/installation.rst
```diff
@@ -65,8 +65,8 @@ Installing |pbm| Using ``apt``
 * Debian 9 ("stretch")
```
EvgeniyPatlan, line 65: also Debian 10 ("buster") is supported

PR #384, opened 2020-06-26
PBM-477 Described PBM upgrade steps

nastena1606:


doc/source/upgrading.rst
```diff
@@ -0,0 +1,143 @@
...
+1. Enable |percona| repository.
+#. Stop |pbm-agent|. 
+#. Install new version packages (the old ones are automatically removed).
+#. Start |pbm-agent|.
+   
+.. _important_notes:   
+
+.. rubric:: Important notes
+ 
+1. Backward compatibility between data backup and restore is supported for upgrades within one major version only (e.g. from 1.1.0 to 1.2.0). When you upgrade |PBM| over several major versions (e.g. from 1.0.0 to 1.2.0), we recommend to make a backup right after the upgrade.
```
akira-kurogane, line 24:
```suggestion
1. Backward compatibility between data backup and restore is supported for upgrades within one major version only (e.g. from 1.1.x to 1.2.y). When you upgrade |PBM| over several major versions (e.g. from 1.0.x to 1.2.y), we recommend to make a backup right after the upgrade.
```

akira-kurogane: The minor version number is not meant to be an issue. Eg. 1.1.0 to 1.2.4, no problem. Likewise 1.1.3 to 1.2.0.

PR #395, opened 2020-07-27
PBM-237 PITR description

nastena1606:


doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,81 @@
...
+   $ pbm config --set pitr.enabled=true
+
+Incremental backups
+================================================================================
+
+When |PITR| is enabled, |pbm-agent| starts slicing the :term:`oplog <Oplog>`. The |pbm-agent| to do the slicing is elected in the same way as secondary nodes elect a new primary. (Find more information in :ref:`pbm.architecture.agent`.)
+
+By default, a slice covers a 10 minute span of oplog events. It can be shorter if |PITR| is disabled or if slicing is stopped by a backup operation.
+
+The oplog slices are stored in the :file:`pbm.Pitr` subdirectory in the :ref:`remote storage defined in the config <pbm.config.initialize>`. To ensure continuous time line, each oplog slice has the start time equal to the end time of the previous oplog or a backup.
```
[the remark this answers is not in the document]
dAdAbird: _"To ensure continuous time line, each oplog slice has the start time equal to the end time of the previous oplog or a backup. "_ - I'm not sure this info is useful here. It's kinda more the internals.
I'd rather pointed out that chunks name reflects the start and end time that chunk covers

doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,84 @@
...
+.. note::
+
+   If you just enabled |PITR|, the time range list in the ``pbm list`` output is empty. It requires 10 minutes for the first chunk to appear in the list.  
+
+Restore to the point in time
+============================
+
+.. important::
+
+   Since a restore operation breaks the time line of oplog slicing, you must disable |PITR| before restoring a database:
```
dAdAbird, line 47: It is not critical, a and I don't know if should mention that, but anyway: the remark that the restore breaks up the timeline is more related to the resumption of the PITR slicing (necessity of the backup after the restore). PITR slicing and restore are not compatible just because the PITR slicer during the restore would capture restore events and put it into the "backup" which is illogical and utterly useless (if not dangerous to the data integrity).

doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,84 @@
...
+   $ pbm config --set=pitr.enabled=true
+
+
+.. rubric:: Delete a backup
+
+When you :ref:`delete a backup <pbm.backup.delete>`, all oplog slices that are made after the specified backup or that are older than the specified time will be deleted.  
+
+.. note::
+
+   When |PITR| is enabled, the most recent backup snapshot won't be deleted.
```
dAdAbird, line 82: Just to be clear. All PITR chunks being deleted only with the snapshot they are related to. So if one deletes the snapshot then and only then any chunks related to this snapshot should be deleted. There is no other way to delete a chunk (at least for the time being)

Ramifications of that:
> all oplog slices that are made after the specified backup or that are older than the specified time will be deleted  

if you specify a particular snapshot, PITR chunks that are related to the next snapshot (if such snapshot exists) won't be deleted:
```
---bcp1----bcp2---bcp3---
$ pbm delete-backup bcp2
---bcp1----bcp3---
```
\* `bcp<N>` - is a snapshot, `-` - oplog slice

again, if we specify a time - some chunks that a younger than that time but related to the deleted snapshots would be deleted as well
```
---bcp1----bcp2------bcp3---
                 ^ - t1
$ pbm delete-backup --older-than t1
---bcp1----bcp3---
```
\* all chunks after bcp2 and up to bcp3 would be deleted

> When |PITR| is enabled, the most recent backup snapshot won't be deleted.

Any chunks related to that backup won't be deleted either

akira-kurogane: > There is no other way to delete a chunk (at least for the time being)

Well, you can delete the files in the pitrChunks/ subdirectory in the remote storage yourself.
I make a point of mentioning this because a user might turn on PITR, discover that it takes up a huge amount of space because write churn is large, disable it, and want to just keep snapshots.

doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,84 @@
...
+|PITR| is available in |PBM| starting from v1.3.0. |PITR| is enabled via the ``pitr.enabled`` config option.
+
+.. code-block:: bash
+
+   $ pbm config --set pitr.enabled=true
+
+Incremental backups
+================================================================================
+
+When |PITR| is enabled, |pbm-agent| starts slicing the :term:`oplog <Oplog>`. The |pbm-agent| to do the slicing is elected in the same way as secondary nodes elect a new primary. (Find more information in :ref:`pbm.architecture.agent`.)
```
akira-kurogane, line 17:
```suggestion
When |PITR| is enabled, |pbm-agent| periodically saves consecutive slices of the :term:`oplog <Oplog>`. A method similar to the way replica set nodes elect a new primary is used to select the |pbm-agent| that saves the oplog slices. (Find more information in :ref:`pbm.architecture.agent`.)
```

doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,84 @@
...
+
+A slice covers a 10 minute span of oplog events. It can be shorter if |PITR| is disabled or if slicing is stopped by a backup operation.
```
akira-kurogane, line 19:
```suggestion
A slice covers a 10 minute span of oplog events. It can be shorter if |PITR| is disabled or interrupted by the start of a backup snapshot operation.
```

doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,84 @@
...
+  	2020-07-14T10:44:44Z
+  	2020-07-14T14:26:20Z
+  	2020-07-17T16:46:59Z
+   PITR <on>:
+  	2020-07-14T14:26:40 - 2020-07-16T17:27:26
+  	2020-07-17T16:47:20 - 2020-07-17T16:57:55
+
+.. note::
+
+   If you just enabled |PITR|, the time range list in the ``pbm list`` output is empty. It requires 10 minutes for the first chunk to appear in the list.
```
akira-kurogane, line 40: :+1:

doc/source/point-in-time-recovery.rst
akira-kurogane: So this should say:
"Since a restore operation breaks the time line of oplog updates |PITR| will be automatically disabled as the restore starts."
right @dAdAbird ?

nastena1606: As far as I understand PITR must be disabled manually. It's not done automatically.

doc/source/point-in-time-recovery.rst
dAdAbird: @akira-kurogane yes good point. But I'd recommend running `pbm config --force-resync` after deleting files on the remote storage to avoid the misleading output of `pbm list`

doc/source/point-in-time-recovery.rst
dAdAbird: Yes, it should be disabled manually, no automation here. My comment was rather a terminology nitpicking.

doc/source/point-in-time-recovery.rst
akira-kurogane: Got it.

Anyhow this is comments on comments, I don't suggest an edit to the *.rst file content above.

doc/source/point-in-time-recovery.rst
akira-kurogane: OK, so it should still be disabled manually. So the instructions just below this line are correct.

doc/source/point-in-time-recovery.rst
```diff
@@ -0,0 +1,85 @@
...
+  	2020-07-17T16:47:20 - 2020-07-17T16:57:55
+
+.. note::
+
+   If you just enabled |PITR|, the time range list in the ``pbm list`` output is empty. It requires 10 minutes for the first chunk to appear in the list.  
+
+Restore to the point in time
+============================
+
+A restore and oplog updates are incompatible operations and cannot be run simultaneously. You must disable |PITR| before restoring a database:
```
akira-kurogane, line 45:
```suggestion
A restore and |PITR| incremental backups are incompatible operations and cannot be run simultaneously. You must disable |PITR| before restoring a database:
```

akira-kurogane: Along with the edit suggestion above I realized something. The "Restoring a backup" section in running.rst should also advise that users should disable PITR first before running pbm restore.

PR #412, opened 2020-09-04
PBM-549 Removed Contact us section

nastena1606:
Modified index page


akira-kurogane (approved):
I assume this was a request from elsewhere. Given that, I don't see a problem.

PR #417, opened 2020-09-22
PBM-485 fix: Upload 100gb+ db on aws

dAdAbird:
MaxUploadParts is 1e4 so with PartSize 10Mb the max allowed file size would be ~ 97.6Gb. Hence if the file size is bigger we're enlarging PartSize so PartSize * MaxUploadParts could fit the file. If calculated PartSize is smaller than the default we leave the default.
If UploadPartSize option was set we use it instead of the default. Even with the UploadPartSize set the calculated PartSize woulbe used if it's bigger.

If backup won't be compressed we're assuming that the dump size could be up to 4x lagrer due to mongo's wieredtiger compression. Otherwise, we assume that it would 
be covered by our compression. But in any case, there is an extra 10% gap while calculating chunk size.

### DIFF
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -207,8 +207,20 @@ func (b *Backup) run(bcp pbm.BackupCmd) (err error) {
 		return errors.Wrap(err, "set shard's first write ts")
 	}
 
+	// Get size of the largest db in replset
+	// so the S3 uploader can calculate proper chunk size.
+	//
+	// In fact, for all DBs the upload size would be the same
+	// and equals to the size of the largest DB. We're ok with that
+	// since there are no easy way to define (distinct) DBs in upload
+	// func given we're piping dump thru the mongodump util.
+	sz, err := b.node.MaxDBSize()
+	if err != nil {
+		return errors.Wrap(err, "mongodump")
+	}
+
 	dump := newDump(b.node.ConnURI(), runtime.NumCPU()/2)
-	_, err = Upload(b.ctx, dump, stg, bcp.Compression, rsMeta.DumpName)
+	_, err = Upload(b.ctx, dump, stg, bcp.Compression, rsMeta.DumpName, sz)
 	if err != nil {
 		return errors.Wrap(err, "mongodump")
 	}
@@ -252,7 +264,7 @@ func (b *Backup) run(bcp pbm.BackupCmd) (err error) {
 	}
 
 	oplog.SetTailingSpan(oplogTS, lwTS)
-	_, err = Upload(b.ctx, oplog, stg, bcp.Compression, rsMeta.OplogName)
+	_, err = Upload(b.ctx, oplog, stg, bcp.Compression, rsMeta.OplogName, -1)
 	if err != nil {
 		return errors.Wrap(err, "oplog")
 	}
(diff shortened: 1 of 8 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

pbm/backup/backup.go
```diff
@@ -252,7 +264,7 @@ func (b *Backup) run(bcp pbm.BackupCmd) (err error) {
 	}
 
 	oplog.SetTailingSpan(oplogTS, lwTS)
-	_, err = Upload(b.ctx, oplog, stg, bcp.Compression, rsMeta.OplogName)
+	_, err = Upload(b.ctx, oplog, stg, bcp.Compression, rsMeta.OplogName, -1)
```
akira-kurogane, line 267: Is this leaving PBM at risk of failing if an oplog segment is > 100GB?
It's reasonable not to expect this i.m.o., but I would add a comment such as 'Not setting size parameter in Upload() because we don't expect an oplog span to ever be bigger than the default AWS max file size.'

pbm/node.go
```diff
@@ -106,6 +106,28 @@ func (n *Node) GetInfo() (*NodeInfo, error) {
...
+		} `bson:"databases"`
+	}{}
+	err := n.cn.Database(DB).RunCommand(n.ctx, bson.D{{"listDatabases", 1}}).Decode(i)
+	if err != nil {
+		return 0, errors.Wrap(err, "run mongo command listDatabases")
+	}
+
+	sized := 0
+	for _, db := range i.Databases {
+		if db.SizeOnDisk > sized {
```
akira-kurogane, line 124: Proportional to the amount of WT blocks for deleted documents that might exist in the data files this may be larger than the dumped file. But other than that this is will be an accurate prediction of the dump size with s2 compression.

pbm/storage/s3/s3.go
```diff
@@ -121,9 +123,23 @@ func (s *S3) Save(name string, data io.Reader) error {
...
 		}
 
+		// MaxUploadParts is 1e4 so with PartSize 10Mb the max allowed file size
+		// would be ~ 97.6Gb. So if the file size is bigger we're enlarging PartSize
+		// so PartSize * MaxUploadParts could fit the file.
+		// If calculated PartSize is smaller than the default we leave the default.
+		// sizeb usually would be an uncompressed db size on disk
+		partSize := defaultPartSize
+		if sizeb > 0 {
+			ps := sizeb / s3manager.MaxUploadParts * 99 / 100 // shed 1% just to be sure we fit in even when data is not compressed
```
akira-kurogane, line 133: Ah, wait. The SizeOnDisk property fetched in MaxDBSize() is the size of the *.wt files, but the WiredTiger storage engine has compressed their document data using s2 already.

I think we my need to do something such as: multiply by a factor if zip or no compression type is used.

pbm/node.go
akira-kurogane: To get the uncompressed size we'd have to run a dbStats command and look at the "dataSize" property. This has to be called database-by-database.

pbm/node.go
```diff
@@ -106,6 +106,28 @@ func (n *Node) GetInfo() (*NodeInfo, error) {
 	return i, nil
 }
 
+// MaxDBSize returns the size of the largest database on replicaset.
+// It is the of files on disk in bytes
```
akira-kurogane, line 110: Why just the largest db? When we use mongodump with --archive all the database namespaces go into one file.

dAdAbird: Yep, my glitch

JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):
LGTM

PR #418, opened 2020-09-24
PBM-547 fix: move delete ops to agent

dAdAbird:
Running deletes by “pbm” CLI might lead to issues with file permissions (file ower/group) on FS. So delete-backup operations should be executed by agents.

1. Locking

In order to run delete operations agents have to acquire locks (to decide which one is going to do the job). This means that no deletes allowed during:
1. Restores. Which is good since the user won’t able to delete backup from which restore is performing. Now that’s possible.
2. Backups. Which gives nothing since delete op refuses to do anything until the backup is either in “done” or “error” state. So the user can’t shoot in their foot deleting running backup even without locks.
3. PITR. Which again brings no value since delete op has the protection against deleting latest snapshot and derived from it pirt slices if PITR is ON.
4. Resync. Which is kinda “who cares”. If backup already synced its ok to delete it even while the other is still syncing.

So locks on delete op protect only from really bonker actions during the restore. But it’s going to really nag if PITR is running whereas user needs faffing around turning PITR off to delete something and then turning it on. No good at all.
Adding extra magic complexity to sync delete with pitr (as we do with backups during pitr) doesn’t worth it in this case.

This PR introduces non-exclusive operations' locks. Meaning that whenever such operation is running there is shouldn't be any other operation of the same type running concurrently, but the other operation is allowed. E.g. there are shouldn't be running more than one delete-backup operation but delete-backup may run along with the PITR slicing.

2. Async CLI

Having delete ops running by “pbm” CLI we have synchronous feedback to the user. Moving it to agents and yet not having status command we lose that interaction, by meaning that user has to investigate agent logs to find out if any error happened or has operation finished yet. 
To preserve synchronous feedback “pbm” CLI will wait up to 60 sec for the operation to be finished by an agent. And notify the user if the delete was finished or any execution errors happened during this timeframe. If in 60 sec operation is still in progress, CLI notify the user and return control.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -145,7 +145,8 @@ func (a *Agent) Delete(d pbm.DeleteBackupCmd) {
 		}
 		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "done")
 	default:
-		a.log.Error(pbm.CmdDeleteBackup, "", "malformed command: %v", d)
+		a.log.Error(pbm.CmdDeleteBackup, "", "malformed command received in Delete() of backup: %v", d)
+
 	}
 }
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -128,18 +128,19 @@ func (a *Agent) Delete(d pbm.DeleteBackupCmd) {
 	case d.OlderThan > 0:
 		t := time.Unix(d.OlderThan, 0).UTC()
 		tstr := t.Format("2006-01-02T15:04:05Z")
-		a.log.Info(pbm.CmdDeleteBackup, tstr, "deleteing backups older than %v", t)
+		a.log.Info(pbm.CmdDeleteBackup, tstr, "deleting backups older than %v", t)
 		err := a.pbm.DeleteOlderThan(t)
 		if err != nil {
-			a.log.Error(pbm.CmdDeleteBackup, tstr, "deleteing: %v", err)
+			a.log.Error(pbm.CmdDeleteBackup, tstr, "deleting: %v", err)
 			return
 		}
 		a.log.Info(pbm.CmdDeleteBackup, tstr, "done")
 	case d.Backup != "":
-		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "deleteing backup")
+		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "deleting backup")
 		err := a.pbm.DeleteBackup(d.Backup)
 		if err != nil {
-			a.log.Error(pbm.CmdDeleteBackup, d.Backup, "deleteing: %v", err)
+			a.log.Error(pbm.CmdDeleteBackup, d.Backup, "deleting: %v", err)
+
 			return
 		}
 		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "done")
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -91,11 +91,22 @@ func (a *Agent) Start() error {
 
 // Delete deletes backup(s) from the store and cleans up its metadata
 func (a *Agent) Delete(d pbm.DeleteBackupCmd) {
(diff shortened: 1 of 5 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

JNKPercona:
| Test name  | Status |
| mongodb 4.0 | failed |
| mongodb 3.6 | failed |
| mongodb 4.2 | failed |

agent/agent.go
```diff
@@ -87,6 +89,65 @@ func (a *Agent) Start() error {
...
+		if err != nil {
+			a.log.Error(pbm.CmdDeleteBackup, "", "release lock: %v", err)
+		}
+	}()
+
+	switch {
+	case d.OlderThan > 0:
+		t := time.Unix(d.OlderThan, 0).UTC()
+		tstr := t.Format("2006-01-02T15:04:05Z")
+		a.log.Info(pbm.CmdDeleteBackup, tstr, "deleteing backups older than %v", t)
```
akira-kurogane, line 131:
```suggestion
		a.log.Info(pbm.CmdDeleteBackup, tstr, "deleting backups older than %v", t)

```

agent/agent.go
```diff
@@ -87,6 +89,65 @@ func (a *Agent) Start() error {
...
+		if err != nil {
+			a.log.Error(pbm.CmdDeleteBackup, tstr, "deleteing: %v", err)
+			return
+		}
+		a.log.Info(pbm.CmdDeleteBackup, tstr, "done")
+	case d.Backup != "":
+		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "deleteing backup")
+		err := a.pbm.DeleteBackup(d.Backup)
+		if err != nil {
+			a.log.Error(pbm.CmdDeleteBackup, d.Backup, "deleteing: %v", err)
```
akira-kurogane, line 142:
```suggestion
		a.log.Info(pbm.CmdDeleteBackup, tstr, "deleting backups older than %v", t)
		err := a.pbm.DeleteOlderThan(t)
		if err != nil {
			a.log.Error(pbm.CmdDeleteBackup, tstr, "deleting: %v", err)
			return
		}
		a.log.Info(pbm.CmdDeleteBackup, tstr, "done")
	case d.Backup != "":
		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "deleting backup")
		err := a.pbm.DeleteBackup(d.Backup)
		if err != nil {
			a.log.Error(pbm.CmdDeleteBackup, d.Backup, "deleting: %v", err)

```

agent/agent.go
```diff
@@ -87,6 +89,66 @@ func (a *Agent) Start() error {
...
+		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "deleting backup")
+		err := a.pbm.DeleteBackup(d.Backup)
+		if err != nil {
+			a.log.Error(pbm.CmdDeleteBackup, d.Backup, "deleting: %v", err)
+
+			return
+		}
+		a.log.Info(pbm.CmdDeleteBackup, d.Backup, "done")
+	default:
+		a.log.Error(pbm.CmdDeleteBackup, "", "malformed command: %v", d)
```
akira-kurogane, line 148:
```suggestion
		a.log.Error(pbm.CmdDeleteBackup, "", "malformed command received in Delete() of backup: %v", d)

```

akira-kurogane (approved):
LGTM

JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

PR #421, opened 2020-09-28
PBM-537 feature: check oplog sufficient range

dAdAbird:
During the oplog slicing, before processing the first oplog record we check if oplog has sufficient range,
i.e. if there are no gaps between the ts of the last backup or slice and
the first record of the current slice. Whereas the request is >= last_saved_ts
and no monotonic increment for oplog records we can't tell
if the returned oldest record is the first since last_saved_ts or
there were other records that are now removed because of oplog collection capacity.
So after we retrieved the first record of the current slice we check if there is
at least one preceding record (basically if there is still record(s) from the previous set).
If so, we can be sure we have a contiguous history with respect to the last_saved_slice.
There's a chance of false-negative though.

We should do this check only after we retrieved the first record of the set. Otherwise,
there is a possibility some records would be erased in a time span between the check and first
recorded retrieval due to ongoing write traffic (i.e. oplog append). 

In case of error `pbm list` will display
```
...
!Failed to run PITR backup. Agent logs:
  rs2: 2020-09-29T20:07:23.000+0300 [ERROR] pitr: streaming oplog: unable to upload chunk 1601397333.1601399234: read data: oplog has insufficient range, some records since the last saved ts {1601397333 1} are missing. Run `pbm backup` to create a valid starting point for the PITR.
```

https://jira.percona.com/browse/PBM-537

### DIFF
--- a/pbm/backup/oplog.go
+++ b/pbm/backup/oplog.go
@@ -38,7 +38,7 @@ type ErrInsuffRange struct {
 }
 
 func (e ErrInsuffRange) Error() string {
-	return fmt.Sprintf("oplog has insufficient range, not enough data from starting point %v", e.t)
+	return fmt.Sprintf("oplog has insufficient range, not enough data from starting point %v. Run `pbm backup` to create a valid starting point for the PITR", e.t)
 }
 
 // WriteTo writes an oplog slice between start and end timestamps into the given io.Writer
@@ -80,7 +80,7 @@ func (ot *Oplog) WriteTo(w io.Writer) (int64, error) {
 		}
 		// Before processing the first oplog record we check if oplog has sufficient range,
 		// i.e. if there are no gaps between the ts of the last backup or slice and
-		// the first record of the current slice. Whereas the request is >= last_saved_ts
+		// the first record of the current slice. Whereas the request is ">= last_saved_ts"
 		// we don't know if the returned oldest record is the first since last_saved_ts or
 		// there were other records that are now removed because of oplog collection capacity.
 		// So after we retrieved the first record of the current slice we check if there is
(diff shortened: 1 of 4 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

pbm/backup/oplog.go
```diff
@@ -62,13 +71,37 @@ func (ot *Oplog) WriteTo(w io.Writer) (int64, error) {
...
+		// So after we retrieved the first record of the current slice we check if there is
+		// at least one preceding record (basically if there is still record(s) from the previous set).
+		// If so, we can be sure we have a contiguous history with respect to the last_saved_slice.
+		//
+		// We should do this check only after we retrieved the first record of the set. Otherwise,
+		// there is a possibility some records would be erased in a time span between check and first
+		// recorded retrieval due to ongoing write traffic (i.e. oplog append). There's a chance of
+		// false-negative though.
+		if !rcheck {
+			c, err := cl.CountDocuments(ctx, bson.M{"ts": bson.M{"$lte": ot.start}})
```
akira-kurogane, line 95: If a count command is run it will scan all documents (or least those before the timestamp).

I suggest doing a findOne() on the same clause and that way the db server can stop scanning oplog documents as soon as possible.

dAdAbird: Eah, I forgot to add option `{limit: 1}` to this call. I believe it does the same - stops scanning docs after the firs find

JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

pbm/backup/oplog.go
dAdAbird: fixed 95b46c1

JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

pbm/backup/oplog.go
akira-kurogane: I didn't realize there was a limit argument for count, but there we are. OK.

akira-kurogane (approved):
LGTM

PR #422, opened 2020-09-30
PBM-547 fix locking

dAdAbird:
The delete op could be finished fast enough so the other node in the RS may acquire a lock for the same operation.
It may lead to errors in logs (and in CLI) "Error: deleting: unable to delete backup in ' ' state".
Meaning the tries to delete the backup that is already has been deleted.
To prevent this races the node that finished delete has to wait some time so the lock should be held at least for 5 sec. Hence other nodes could observe this lock and abort.


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):

PR #427, opened 2020-10-05
PBM-426 Described SSE options

nastena1606:
Added server-side encryption options to the config options section
Extended config file example
Added a note about SSE support in PBM 1.3.2

### DIFF
--- a/doc/source/glossary.rst
+++ b/doc/source/glossary.rst
@@ -79,4 +79,4 @@
      This is the storage that is built on the :term:`S3 <Amazon S3>` API.
  
   Server-side encryption
-     Server-side encryption is the encryption of data on the remote storage. The data is encrypted when it is written to S3 bucket and decrypted when you access the data. 
+     Server-side encryption is the encryption of data by the remote storage server as it receives it. The data is encrypted when it is written to S3 bucket and decrypted when you access the data. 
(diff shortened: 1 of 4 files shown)


doc/source/glossary.rst
```diff
@@ -77,4 +77,6 @@
 
   S3 compatible storage   
      This is the storage that is built on the :term:`S3 <Amazon S3>` API.
-   
\ No newline at end of file
+ 
+  Server-side encryption
+     Server-side encryption is the encryption of data on the remote storage. The data is encrypted when it is written to S3 bucket and decrypted when you access the data.
```
akira-kurogane, line 82: I think this first sentence might be interpreted as saying PBM can do encryption as it saves to the remote storage (including filesystem type).

I'm not sure what to say - does the following seem better?
```suggestion
     Server-side encryption is the encryption of data by the remote storage server as it receives it. The data is encrypted when it is written to S3 bucket and decrypted when you access the data. 
```

akira-kurogane (changes requested):

doc/source/glossary.rst
nastena1606: Yes, Akira. Thank you for your suggestion

akira-kurogane (approved):

dAdAbird (approved):

PR #429, opened 2020-10-05
PBM-562 fit: PITR timeline merging

dAdAbird:
Merging of the PITR timelines from replicates relies on the assumption that RSs are making progress and stops altogether. And if some node fails and then resumes it'll catchup. I.e. there is no gap in timelines possible. But that is the false assumption.
In fact, we can have a station when one more RS stopped slicing while another node(s) making progress. And then after the backup, all nodes resume slicing starting from the backup's end_time leaving the gap in timelines on the RS that had been failed.

This commit changes the algorithm from comparing only timelines with the same index to grinding thru all timelines on other replica sets searching for overlapping ones.

### DIFF
--- a/pbm/pitr.go
+++ b/pbm/pitr.go
@@ -208,6 +208,11 @@ func gettimelines(slices []PITRChunk) (tlines []Timeline) {
 	return tlines
 }
 
+type tlineMerge struct {
+	tl    Timeline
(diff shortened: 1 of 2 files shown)


akira-kurogane (approved):
LGTM

PR #431, opened 2020-10-06
PBM-559 Documented restore config options

nastena1606:
Added a subsection for restoring large backups

### DIFF
--- a/doc/source/running.rst
+++ b/doc/source/running.rst
@@ -98,7 +98,7 @@ Running |pbm.app| Commands
 Configuring a Remote Store for Backup and Restore Operations
 --------------------------------------------------------------------------------
 
-This must be done once, at installation or re-installaton time, before backups can
+This must be done once, at installation or re-installation time, before backups can
 be listed, made, or restored. Please see :ref:`pbm.config`.
 
 .. _pbm.running.backup.listing:
@@ -175,7 +175,7 @@ restore.
    will perform a full all-databases, all collections restore and does not
    offer an option to restore only a subset of collections in the backup, as
    MongoDB's mongodump tool does. But to avoid surprising mongodump users |pbm|
-   as of now (versions 1.x) replicates mongodump's behaviour to only drop
+   as of now (versions 1.x) replicates mongodump's behavior to only drop
    collections in the backup. It does not drop collections that are created new
    after the time of the backup and before the restore. Run a db.dropDatabase()
    manually in all non-system databases (i.e. all databases except "local",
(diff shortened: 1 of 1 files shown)


doc/source/running.rst
```diff
@@ -184,8 +184,25 @@ restore.
 
 .. include:: .res/code-block/bash/pbm-restore-mongodb-uri.txt
 
-After a cluster's restore is complete all mongos nodes will need to be
-restarted to reload the sharding metadata.
+After a cluster's restore is complete, restart all ``mongos`` nodes to reload the sharding metadata.
+
+.. rubric:: Restoring a large backup
+
+When you restore a backup that is large for the currently available memory on the server, the restore operation may fail with the out of memory error. To avoid that error, you can adjust the memory consumption during restore. Define the following options in the |PBM| config:
```
dAdAbird, line 191: Current phrasing sets user expectations to the error which could happen only on the setups with tight memory restrictions. I'd rather put that in a way that "these new options are introduced for better adjustments in environments with tight memory bounds". Also maybe worth to mention that defaults were changed to fit setups with 1Gb and less for the agent.

dAdAbird: @akira-kurogane what do think? ^^^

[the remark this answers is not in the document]
akira-kurogane: Yes, it's not how big the backup is - the database could be 10GB or 10TB. The issue that happened was caused by the amount of memory that will be in use by pbm-agents _at any given instant_ vs. the RAM available. It seems that could be as high as 2GB, but you can tune the rate of IO to be lower with this options in this patch, and that really reduces the instantaneous memory needs. And thus avoid OOM death.

The wording "these new options are introduced for better adjustments in environments with tight memory bounds" is good i.m.o.

akira-kurogane: Seems to be the same comment repeated three times. To cut-and-paste what I replied to the first comment:

_Yes, it's not how big the backup is - the database could be 10GB or 10TB. The issue that happened was caused by the amount of memory that will be in use by pbm-agents_ at any given instant _vs. the RAM available. It seems that could be as high as 2GB, but you can tune the rate of IO to be lower with this options in this patch, and that really reduces the instantaneous memory needs. And thus avoid OOM death._

_The wording "these new options are introduced for better adjustments in environments with tight memory bounds" is good i.m.o._

dAdAbird: yep, it duplicated

dAdAbird (approved):

akira-kurogane (approved):

PR #434, opened 2020-10-07
PBM-568 feature: print s3.uploadPartSize to log

dAdAbird:
Since s3.uploadPartSize is automatically adjusted it could be
useful to see what the value happened to be at the end of the day

Log entry looks like
```
2020-10-07T19:32:28.543+0300 [INFO] s3.uploadPartSize is set to 10485760 (~10Mb)
```

https://jira.percona.com/browse/PBM-568

### DIFF
--- a/pbm/storage/s3/s3.go
+++ b/pbm/storage/s3/s3.go
@@ -1,12 +1,15 @@
 package s3
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


akira-kurogane (approved):
Nice that the log message is being added.

PR #447, opened 2020-10-27
Make it possible to configure number of parallel collections to dump

juusom:
PBM currently dumps a maximum of `runtime.NumCPU()/2` collections concurrently. In our tests, depending on the used compression algorithm, this leads to either saturating the uplink of the backup nodes or very high CPU loads. We have some use cases where applications read from secondaries and we'd prefer to sacrifice some backup speed in order not to compromise those use cases.

This PR introduces a new optional pbm-agent environment variable, `PBM_DUMP_PARALLEL_COLLECTIONS`, for controlling the parallelism. It's not perfect as it doesn't in any way guarantee fairness of resource/bandwidth use, but at least it allows leaving more CPU cores free for other tasks and has an effect on bandwidth usage.

I figured an environment variable would make sense as opposed to using `pbm config` as it's possible to have a cluster with heterogenous hardware, and therefore the configuration should be per-agent.

For the speed tests I hardcoded the value to 1 as they don't seem to use mongodump, and they seem to operate on just one collection anyway.

I don't mind updating the docs if the idea gets approved, but didn't want to do it before.

BTW, `agent/agent.go` has CRLF line terminators while others don't. My IDEA is very eager to fix it but I don't know about your style policies so I left it as it is in order to avoid a horrible diff :)

### DIFF
--- a/cmd/pbm-agent/main.go
+++ b/cmd/pbm-agent/main.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/alecthomas/kingpin"
@@ -21,6 +23,7 @@ func main() {
 		pbmAgentCmd = pbmCmd.Command("run", "Run agent").Default().Hidden()
 
 		mURI = pbmAgentCmd.Flag("mongodb-uri", "MongoDB connection string").Envar("PBM_MONGODB_URI").Required().String()
+		dumpConns = pbmAgentCmd.Flag("parallel-collections", "Number of collections to dump in parallel").Envar("PBM_DUMP_PARALLEL_COLLECTIONS").Default(strconv.Itoa(runtime.NumCPU()/2)).Int()
 
 		versionCmd    = pbmCmd.Command("version", "PBM version info")
 		versionShort  = versionCmd.Flag("short", "Only version info").Default("false").Bool()
@@ -46,10 +49,10 @@ func main() {
 		return
 	}
 
-	log.Println(runAgent(*mURI))
+	log.Println(runAgent(*mURI, *dumpConns))
 }
 
-func runAgent(mongoURI string) error {
+func runAgent(mongoURI string, dumpConns int) error {
 	mongoURI = "mongodb://" + strings.Replace(mongoURI, "mongodb://", "", 1)
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -61,7 +64,7 @@ func runAgent(mongoURI string) error {
 	}
 
 	agnt := agent.New(pbmClient)
-	err = agnt.AddNode(ctx, mongoURI)
+	err = agnt.AddNode(ctx, mongoURI, dumpConns)
 	if err != nil {
 		return errors.Wrap(err, "connect to the node")
 	}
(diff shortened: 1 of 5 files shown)


akira-kurogane (approved):
Having a tuning option for the export rate is good. I have reservations about the naming, and place the option is set in. But after a bit of reflection they don't prevent taking it as it is.

Regarding place - I think it should be in the PBM global config. But then I think this patch's method of adding as env var gives us a way to _experimentally_ try this throttle method out in the field. We can make it a global config val later and deprecate this pbm-agent-only option without much UI disturbance.

Alternative ways of throttling exist. One would be set a max CPU ratio (by some mechanism that ultimately uses setrlimit(RLIMIT_CPU, ...). Another is the niceness level. Another would be to limit the number of CPU cores it can be allowed to run on (probably by using cpuset).

Regarding name, in mongodump the option is "--numParallelCollections". For the goal of having UI that avoids memorizing some new arbitrary info this is the precedent related to this option to look at i.m.o. But given we have 'lowercase-hyphen-joined-name' already for the other PBM command-line args and I expect we'll deprecate it later well let's not worry about it.

Approved from me too @dAdAbird

juusom:
Thanks for the review @akira-kurogane!

I figured the option should be a per-agent one because in theory replica sets / clusters can have heterogenous hardware, even though it might not exactly be advisable. But now that I think of it, in such a case it should be fine to set the global setting according to the smallest node in the cluster.

I was scratching my head with the naming. If this eventually gets moved to the global config, I guess there could be a section for mongodump options that could be named more similarly to the originals.

About throttling, alternatives certainly exist but throttling pbm-agent alone might not do the trick. I'm not too familiar with the performance characteristics but I would guess that with compression disabled, the bottleneck would be either `mongod`'s ability to push out data or the outbound network link. This option lets you, to some extent, control how much load you put on mongo itself, and indirectly to the outbound network.

akira-kurogane:
> I was scratching my head with the naming. If this eventually gets moved to the global config, I guess there could be a section for mongodump options that could be named more similarly to the originals.

Yes, we don't even have a section for it yet. There are the _storage_ options for location, auth, compression from the start of course. It was only recently the _pitr.enabled_ and _restore.batchSize_ and _restore.numInsertionWorkers_ options were added. Those two _restore_ options avoid decompression-related memory usage being so high that small servers (think 2GB or less) users were observing OOM-kills.

In this case it would be a "backup" section I presume. _backup.throttle_net_out_ maybe? Or _backup.throttle_cpu_. But I suggest we see how it goes out in the field before deciding that.

cmd/pbm-agent/main.go
```diff
@@ -21,6 +23,7 @@ func main() {
 		pbmAgentCmd = pbmCmd.Command("run", "Run agent").Default().Hidden()
 
 		mURI = pbmAgentCmd.Flag("mongodb-uri", "MongoDB connection string").Envar("PBM_MONGODB_URI").Required().String()
+		dumpConns = pbmAgentCmd.Flag("parallel-collections", "Number of collections to dump in parallel").Envar("PBM_DUMP_PARALLEL_COLLECTIONS").Default(strconv.Itoa(runtime.NumCPU()/2)).Int()
```
dAdAbird, line 26: I suggest renaming `parallel-collections` flag to `dump-parallel-collections` since there is the same option for the mongorestore which maybe we would expose too

dAdAbird (commented):
See my comments on the flag name. Otherwise LGTM.

BTW, there is the same option for the mongorestore util. It is always set to 1 currently (https://github.com/percona/percona-backup-mongodb/blob/main/pbm/restore/restore.go#L426) And it might be worth having that option adjustable as well (in future). @akira-kurogane @juusom

dAdAbird (approved):

PR #454, opened 2020-11-11
Ignore shard configuration from backup

juusom:
When restoring a backup to a cluster consisting of different nodes than the ones the backup was made from, the shard configuration is overwritten. I don't think it makes sense to restore the shard configuration in any conditions because in any case, before restoring a sharded cluster, you need to have a functional cluster set up to begin with.

I almost had a heart attack the first time I noticed that `sh.status()` on my test cluster claimed to be talking to production shards.

### DIFF
--- a/pbm/restore/restore.go
+++ b/pbm/restore/restore.go
@@ -42,6 +42,7 @@ var excludeFromRestore = []string{
 	"config.locks",
 	"config.system.sessions",
 	"config.cache.*",
+	"config.shards",
 }
 
 type Restore struct {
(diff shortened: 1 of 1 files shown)


akira-kurogane:
Ah, yes. We should avoid restoring _admin.system.version_ collection too. I.i.rc. we already do but I don't know where in the code that is (or isn't) in the code at the moment.

dAdAbird (commented):
LGTM,
and can you also add `admin.system.version` to the exclude list so we would merge it in one go

juusom:
@dAdAbird done

dAdAbird (approved):

akira-kurogane (approved):
Thanks - looks good to me. I appreciate that the second collection has been added in. I almost did that earlier today myself but was worried I'd confuse it for you and Andrew. But all good now.

PR #455, opened 2020-11-11
PBM-555 Retry and reconnect on errors while downloading from S3

dAdAbird:
Currently when PBM downloads and parses the backup file during restore it has to abort if reading or connection error happens. Since file (object) being downloaded and parsed as a whole, retrying won't help much if an error happened after lest's say 500Gb already had been parsed. "Retrying" in such case would mean do it again from the beginning.

This PR addresses issue by downloading S3 objects in chunks using HTTP Range header https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35.  That makes retries possible if an error happened.
Along with the retries, PBM would recreate session with the S3 should conniption being dropped.

https://jira.percona.com/browse/PBM-555

### DIFF
--- a/cmd/pbm/logs.go
+++ b/cmd/pbm/logs.go
@@ -48,8 +48,6 @@ func logs(cn *pbm.PBM) {
 		f = plog.FormatJSON
 	}
 
-	// TODO: need to decouple "logger" and "logs printer"
-	cn.InitLogger("", "")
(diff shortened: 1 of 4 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (commented):
LGTM

PR #457, opened 2020-11-11
PBM-460: Prevent conflicting namespace during the restore

dAdAbird:
Restore should remove tmp user and role collections otherwise
it would be a conflicting namespaces error during restore

This commit addresses issue by
1. Ensuring that restore would try to drop tmp collections even on error.
Which didn't happened before. The drop was run only on success.

2. Backup would ensure there ate no tmp collections and try to drop it otherwise.

There is no way to exclude some collections on mongodump stage
while it accepts "exclude..." option only if sole db specified (¯\_(ツ)_/¯)
So we're trying to clean up on primary node when it gets bcp command
but because there is no guarantee
that primary node reaches that code before some of the secondaries starts backup
we have to double-check in backup.run() before mongodump and fail on the dirty state.

https://jira.percona.com/browse/PBM-460

### DIFF
--- a/agent/snapshot.go
+++ b/agent/snapshot.go
@@ -49,7 +49,28 @@ func (a *Agent) CancelBackup() {
 func (a *Agent) Backup(bcp pbm.BackupCmd) {
 	l := a.log.NewEvent(string(pbm.CmdBackup), bcp.Name)
 
-	q, err := backup.NodeSuits(a.node)
+	nodeInfo, err := a.node.GetInfo()
+	if err != nil {
+		l.Error("get node info: %v", err)
+		return
+	}
+
+	// In case there are some leftovers from the restore.
+	//
+	// There is no way to exclude some collections on mongodump stage
+	// while it accepts "exclude..." option only if sole db specified (¯\_(ツ)_/¯)
+	// So we're trying to clean up on primary node but because there is no guarantee
+	// that primary node reaches this code before some of the secondaries starts backup
+	// we have to double-check in backup.run() before mongodump and fail on dirty state.
+	if nodeInfo.IsPrimary && nodeInfo.Me == nodeInfo.Primary {
+		err = a.node.DropTMPcoll()
+		if err != nil {
+			l.Error("ensure no tmp collections: %v", err)
+			return
+		}
+	}
+
+	q, err := backup.NodeSuits(a.node, nodeInfo)
 	if err != nil {
 		l.Error("node check: %v", err)
 		return
@@ -58,12 +79,6 @@ func (a *Agent) Backup(bcp pbm.BackupCmd) {
 	// node is not suitable for doing backup
 	if !q {
 		l.Info("node is not suitable for backup")
-		return
-	}
-
-	nodeInfo, err := a.node.GetInfo()
-	if err != nil {
-		l.Error("get node info: %v", err)
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

agent/snapshot.go
```diff
@@ -49,7 +49,28 @@ func (a *Agent) CancelBackup() {
...
+	//
+	// There is no way to exclude some collections on mongodump stage
+	// while it accepts "exclude..." option only if sole db specified (¯\_(ツ)_/¯)
+	// So we're trying to clean up on primary node but because there is no guarantee
+	// that primary node reaches this code before some of the secondaries starts backup
+	// we have to double-check in backup.run() before mongodump and fail on dirty state.
+	if nodeInfo.IsPrimary && nodeInfo.Me == nodeInfo.Primary {
+		err = a.node.DropTMPcoll()
+		if err != nil {
+			l.Error("ensure no tmp collections: %v", err)
```
akira-kurogane, line 68: This error message should list what the "temp" collections are. The user won't know.

pbm/backup/backup.go
```diff
@@ -183,6 +183,14 @@ func (b *Backup) run(bcp pbm.BackupCmd) (err error) {
 		return errors.Wrap(err, "waiting for start")
 	}
 
+	// A fallback. If there are some leftovers and main node
+	// hasn't cleaned up it yet we should stop and notify user
+	// since backup would be unrestorable.
+	err = b.node.EnsureNoTMPcoll()
+	if err != nil {
+		return errors.Wrap(err, "EnsureNoTMPcoll")
```
akira-kurogane, line 191:
```suggestion
		return errors.Wrap(err, "Error-exiting from Backup run() because processing check EnsureNoTMPcoll() failed. See preceding messages in log for collection names.")
```

akira-kurogane (changes requested):

pbm/backup/backup.go
dAdAbird: The error here will contain collections name. Now it looks like:
```
backup: EnsureNoTMPcoll: pbm tmp collections exists. It will lead to failed restore. Drop collections: admin.pbmRRoles, admin.pbmRUsers
```

looking at it now I reckon it needs some rephrasing. I suggest:
```
backup: EnsureNoTMPcoll: PBM temporary collections exists. It will lead to a failed restore. Please, drop next collections manually: admin.pbmRRoles, admin.pbmRUsers
```

I don't quite in favour of long phrasing in logs. E.g. `Error-exiting from Backup run() because processing check EnsureNoTMPcoll() failed` - anyway gives nothing valuable to the user, and for programmer `EnsureNoTMPcoll ` is enough to identify the place in the code and it makes less visual noise in the log.

agent/snapshot.go
dAdAbird: this error message contain collection names. in log it would be like:
```
backup: ensure no tmp collections: drop tmp roles collection pbmRRoles: blah-blah-blah
```

dAdAbird: And here is error would happen only if something went wrong during the collections removing (mongo error while Drop or so). We do double-check for the collection in the backup. And that error returns the collection name and asking to delete it manually if such collections left.

akira-kurogane: Ah, so there would be proceeding log messages with the collection name(s).

Ah, sorry. I saw those "drop tmp users collection %s" messages during the first review but I thought they were all within the EnsureNoTMPcoll() function just due to my carelessness.

akira-kurogane (approved):

JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

PR #458, opened 2020-11-12
PBM-555 don't retry download after mongorestore errors

dAdAbird:
Just a chore. If mongorestore had exited and closed pipe before the download is finished it means that some error happened on that side. So there is no reason in retrying download and spamming log with failed attempts.


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):

PR #464, opened 2020-11-24
PBM-556 Introduce operation ID and epoch

dAdAbird:
This PR adds the notion of OPID (Operation ID) and Epoch.

OPID
=====
OPID is a unique ID that being given to any command in the moment of its creation. We use mongo's ObjectID from created document collection pbmCmd. OPID used to identify particular operation thru logs/locks and other artefacts created as the operation progress (or fails) in the system. For example, we could have many runs of the restore command for the very same backup. Despite backup name would be same each command should be definitely distinct from each other as well its trace in the system. Although most of the operations have unique identifiers like "name" backups. It differs from op to op which creates tangle. OPID is the part of every log.Event as well each lock created on behalf of any op except for long-running PITR slicing contains OPID.

Also, PR adds pbmOpLog collection to keep a log of each operation (except for PITR slicing) which helps to have a better retrospective. And preventing having duplicating operations due to lags.

Without such a log we could have a lagging node that starts parsing the command after it already has been done by another node. Lagging node succeeds in acquiring a log (other node had released the lock after success) and starts operation second time. Now to start any op it's not enough to get a lock, but op with unique OPID should be registered in `pbmOpLog` in order to move on.

Epochs
=======
See it as the OPID for long-standing ops - now only PITR slicing. The idea is that operation (that may influence PITR slicing) moves the system from one epoch to another (epoch = current cluster time). Which helps have a better retrospective of the system in respect to the PITR. For example, we could clearly state that the log events from previous epochs are not relevant anymore.
Like OPID Epoch is being a part of each log event, lock and pbmOpLog record.

https://jira.percona.com/browse/PBM-556, also related: https://jira.percona.com/browse/PBM-435


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):

PR #465, opened 2020-12-01
Pbm 590 doc restructure

nastena1606:
[PerconaBackupforMongoDB.pdf](https://github.com/percona/percona-backup-mongodb/files/5621833/PerconaBackupforMongoDB.pdf)
Attaching the pdf version of the documentation for reference


akira-kurogane (approved):
Thanks Anastasia. It looks good, much appreciated.

Akira

sandraromanchenko (commented):
1. Remove 'yum update' part from instructions, since it isn't the same as 'apt-get update'. Command 'apt-get update' is used to download package information from all configured sources, while 'yum update' actually updates all packages on OS, which isn't required for PBM install and might be even harmful sometimes. 
2. Building from source code part: in order to build pbm from source in addition to go/make it's also required to install git. Also to build pbm the command should be either 'make build' or 'make install' instead of 'make' since build will fail due to inability to run tests.
3. Isn't it better to use default 27017 port for PBM_MONGODB_URI in examples? IMHO, it will be less confusing for end-users, otherwise it might create an impression that PBM somehow use some extra port
4. 'Starting a backup with compression section' - it creates an impression that simple 'pbm backup' command creates backup without compression, maybe it's worth renaming it to smth like 'select different compression type for backup'
5. 'Remote backup storage options' section - option storage.type isn't described 
5. No description of pbm-speed-test utility

akira-kurogane:
Hi @sandraromanchenko 

Those 6 items above are all well-spotted good points.

Regarding the port number one ( # 6) though - in a cluster the default port is 27018 for shardsvr nodes and 27019 for configsvr nodes.

qikiqi:
I think [this line](https://github.com/percona/percona-backup-mongodb/blame/d4b3250872245c7bceae799f9909024473d313a2/doc/source/architecture.rst#L59) contains a typo and should say:
```
- *admin.pbmBackups* (Log / status of each backup)
```

Figured I'd rather point it here than create an additional PR just for this.

sandraromanchenko (approved):

PR #466, opened 2020-12-02
PBM-600 Fixed incorrect rendering of a variable

nastena1606:
Reworded run pbm-agent example


patrickbirch (commented):
"and another one - to the configsvr"
(doesn't need the dash)
The mention of versions required me to re-read the paragraph several times. Maybe:
For versions before 1.x, |pbm| performs ...
As of version 1.x and later, |pbm| replicates

patrickbirch (approved):
LGTM

PR #468, opened 2020-12-10
Add config for excluding restore namespaces

juusom:
In addition to the hardcoded list of excluded namespaces, allow configuring more.

My golang-fu isn't strong enough to quickly set up support for generic slice configuration options that you could manipulate via the CLI, so I made this a comma separated string that's parsed where it's used.

The motivation for doing this is to work around https://jira.percona.com/browse/PBM-601 by excluding the problematic collection.

### DIFF
--- a/pbm/config.go
+++ b/pbm/config.go
@@ -52,8 +52,9 @@ type StorageConf struct {
 
 // RestoreConf is config options for the restore
 type RestoreConf struct {
-	BatchSize           int `bson:"batchSize" json:"batchSize,omitempty" yaml:"batchSize,omitempty"` // num of documents to buffer
-	NumInsertionWorkers int `bson:"numInsertionWorkers" json:"numInsertionWorkers,omitempty" yaml:"numInsertionWorkers,omitempty"`
+	BatchSize            int      `bson:"batchSize" json:"batchSize,omitempty" yaml:"batchSize,omitempty"` // num of documents to buffer
+	NumInsertionWorkers  int      `bson:"numInsertionWorkers" json:"numInsertionWorkers,omitempty" yaml:"numInsertionWorkers,omitempty"`
+	ExcludeNamespaces    string   `bson:"excludeNamespaces" json:"excludeNamespaces" yaml:"excludeNamespaces"`
 }
 
 type confMap map[string]reflect.Kind
--- a/pbm/restore/restore.go
+++ b/pbm/restore/restore.go
@@ -410,6 +410,15 @@ func (r *Restore) RunSnapshot() (err error) {
 		numInsertionWorkers = cfg.Restore.NumInsertionWorkers
 	}
 
+	splitFunc := func(c rune) bool {
+		return c == ','
+	}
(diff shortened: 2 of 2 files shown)


dAdAbird (commented):
I reckon ExcludeNamespaces should be a list, so in config it would be:
```yaml
restore:
  excludeNamespaces:
     - name.space1
     - name.space2
```

@akira-kurogane what do you think?

juusom:
Yeah, I agree but didn't have time to figure out how to make it configurable via the CLI interface (I'm far from fluent with Go). I might have time to take another stab at it during the holidays.

akira-kurogane:
> I reckon ExcludeNamespaces should be a list, so in config it would be:
> 
> ```yaml
> restore:
>   excludeNamespaces:
>      - name.space1
>      - name.space2
> ```
> 
> @akira-kurogane what do you think?

That seems the right way to put it into the JSON/YAML structure of the config object, I agree.

I'm having worries about proceeding with adding namespace filters though. There are some users who are keen for PBM to support these as mongodump + mongorestore do. There are also requests to make physical backups (i.e. from WiredTiger snapshots) instead of logical-style backups. The two requests are contradictory requirements because WiredTiger does not (yet, as of v4.4) support manipulation of collections separately - you can only back up the mongod node's entire data directory's worth of data files.

akira-kurogane:
> Yeah, I agree but didn't have time to figure out how to make it configurable via the CLI interface (I'm far from fluent with Go). I might have time to take another stab at it during the holidays.

If you like we'll take care of picking the format that is uniform with the previously developed options.

dAdAbird:
> Yeah, I agree but didn't have time to figure out how to make it configurable via the CLI interface
Oh, I see, sure. I think it should be a list in the config and "key=val,key=val" in the CLA. And the CLA representation would be transformed into the list somewhere in *PBM.SetConfigVar method.

juusom:
> > Yeah, I agree but didn't have time to figure out how to make it configurable via the CLI interface (I'm far from fluent with Go). I might have time to take another stab at it during the holidays.
> 
> If you like we'll take care of picking the format that is uniform with the previously developed options.

Yeah, I certainly don't mind if you do.

Regarding namespace filtering in general, a snapshot based backup/restore sounds like an entirely different procedure than the current one and there will be quite a few issues to sort out in any case - including the already-existing hardcoded excluded collections. IMHO, such fundamentally different ways of backing up and restoring can be expected to support different options.

In our particular case, this is currently the only way I've found to work around PBM-601. More generally, I can imagine there's a good number of use cases where you have collections that contain lots of transient data that doesn't need backing up. In fact, maybe the exclusion should be done in the backup side.

dAdAbird:
Hi @juusom. Sorry for the slow response on our end. It got lost among other tasks. We have a global task to make it possible to choose databases/collections for backup/restore. Witch inclusive or exclusive lists. That will cover this PR. So I'll close the PR. Feel free to reopen and comment if needed. Thanks!

PR #469, opened 2020-12-10
No need to wait and hold a lock for fast ops

dAdAbird:
OPID would protect from running the same op twice by different nodes.

Moreover, waiting may trigger bug in k8s operator.  If the database is relatively small -  backups went really fast. Let’s say in 10sec backup being marked as successfully done. But because of timeouts needed for some internal sync pbm-agent would hold the lock for at least 34sec. So we have a situation when operator observes “backup done” and send the next backup command to the PBM. But since some agents still are holding locks this operation either being discarded by PBM straight ahead (because of timing some of the replset may actually start a backup and then “fail” it because “couldn’t get a response from all shards”).
Anyway, the issue is that for small and fast backup operator might think that backup is done (and it is really done) but the new operation can’t be scheduled while lock(s) is still being held. It is not an issue with the pbm cli since it checks for locks beforehand.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -142,7 +142,6 @@ func (a *Agent) Delete(d pbm.DeleteBackupCmd, opid pbm.OPID, ep pbm.Epoch) {
 		}
 	}()
 
-	tsstart := time.Now()
 	switch {
 	case d.OlderThan > 0:
 		t := time.Unix(d.OlderThan, 0).UTC()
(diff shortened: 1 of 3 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):
Cool. Nice to eliminate some timeout-based 'locks'.

akira-kurogane (approved):
Cool. Nice to eliminate some timeout-based 'locks'.

PR #470, opened 2020-12-11
PBM-602 fix: discard failed PITR chunks

dAdAbird:
Any upload on storage would create the target file first and then will
stream data into it (data could be lager than the available RAM). And if
something went wrong while reading data we may end up with an already created
empty or inconsistent file. PITR chunks have no metadata along with it to
indicate any failed state. Although the failed PITR range won't be saved in
internal db as the available for restore. It would get in there after the storage
resync. So on any upload errors during the PITR slicing, we have to ensure no
leftovers on storage. And also skip empty PITR chunks during resync.

https://jira.percona.com/browse/PBM-602

### DIFF
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -403,11 +403,13 @@ func Upload(ctx context.Context, src Source, dst storage.Storage, compression pb
 	case <-saveDone:
 	}
 
+	r.Close()
+
 	if !err.nil() {
 		return 0, err
 	}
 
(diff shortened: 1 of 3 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):
LGTM

PR #505, opened 2021-02-03
Revert PBM-520: fix node' fails propagation during backup

dAdAbird:
Changes introduced in https://github.com/percona/percona-backup-mongodb/pull/396 broke error propagation during backup. Condition for marking the replset meta with an error was always unreachable since meta status wasn't reread from the DB but rather used the local "init" version. As the result, if one agent failed during the backup it would exit with an error but error state won't be propagated to the backup meta so other agents would be stuck in waiting of failed node to move further (as they hadn't observed the node actually failed).

This commit removes that condition for the replset' meta updates. 

After the introduction of OPID and operations log, there is no need in extra effort to deal with lagging nodes. PBM oplog solves that. (commit: https://github.com/percona/percona-backup-mongodb/commit/9c548cd0c49103915a508698f0e17ae53483dacc)

### DIFF
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -53,15 +53,15 @@ func (b *Backup) Run(bcp pbm.BackupCmd, opid pbm.OPID, l *plog.Event) (err error
 	return b.run(bcp, opid, l)
 }
 
-type errMetaExists struct {
+type errSetMeta struct {
 	error
 }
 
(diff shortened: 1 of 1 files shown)


JNKPercona:
| Test name  | Status |
| mongodb 3.6 | failed |
| mongodb 4.0 | failed |
| mongodb 4.2 | failed |

akira-kurogane (approved):

PR #507, opened 2021-02-07
PBM-636: restore config. preserveUUID

coraxster:
https://jira.percona.com/browse/PBM-636

This change will let to config preserveUUID variable.

Collection UUID in sharded env should be the same at each shard + config server collection info db.
But looks like in some cases we don't need to preserve it.

`preserveUUID=false` commit
https://github.com/percona/percona-backup-mongodb/commit/4844a20876719c0c01d5ead54458b76627475c66#diff-2c9604cc50e20be7992d9ec0a27f57a9bf91a2ea9b2aeb848e2cc1d8983532b2R183

[PBM-419](https://jira.percona.com/browse/PBM-419) `preserveUUID=false` feature branch merge
https://github.com/percona/percona-backup-mongodb/commit/99ae9f5c

### DIFF
--- a/doc/source/configuration-options.rst
+++ b/doc/source/configuration-options.rst
@@ -166,6 +166,7 @@ Backup restore options
    restore:
      batchSize: <int>
      numInsertionWorkers: <int>
+     preserveUUID: <bool>
 
 .. option:: batchSize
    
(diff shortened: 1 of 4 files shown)


dAdAbird:
@coraxster Changes in https://github.com/percona/percona-backup-mongodb/pull/533 fix preserveUUID related issues. So now preserveUUID is always ON and we don't see any circumstance when a user would want to switch it off.

Thank you for your contribution and for raising this issue!

PR #513, opened 2021-03-05
PITR state update on demand

ollevche:
For example, user tries to restore and disables PITR right before that. PBM Agent has not updated PITR state yet and throws an error.

PITR state can be updated on demand before proceeding to restore.


ollevche:
Hey, @dAdAbird,

Thanks for the reply and code comments! 

That's a rough implementation just to show the idea. So some things aren't thought through.

[PBM-662](https://jira.percona.com/browse/PBM-662) is going to help with this.

PR #514, opened 2021-03-10
PBM-634 Doc: added pbm logs and pbm status to commands page

nastena1606:


akira-kurogane (approved):
(For reference PR #527 was where pbm log was added)

Thanks Anastasia, LGTM.

PR #515, opened 2021-03-10
PBM-648 Doc: How to schedule backups using cron

nastena1606:
Added a new How to doc

### DIFF
--- a/doc/source/schedule-backup.rst
+++ b/doc/source/schedule-backup.rst
@@ -1,0 +1,70 @@
+.. _schedule: 
+
+Schedule backups
+==============================
+
+In |PBM| version 1.4.1 and earlier, we recommend using ``crond`` or similar services to schedule backup snapshots.
+
+.. important::
+
+   Before configuring ``crond``, make sure that you have :ref:`installed <install>` and :ref:`configured <initial-setup>` |PBM| to make backups in your database. Start a backup manually to verify this: :command:`pbm backup`. 
+
+The recommended approach is to create a :file:`crontab` file in the :dir:`/ect/cron.d` directory and specify the command in it. This simplifies server administration especially if multiple users have access to it.
+
+|pbm.app| CLI requires a valid MongoDB URI connection string to authenticate in MongoDB. Instead of specifying the MongoDB URI connection string as a command line argument, which is a potential security risk, we recommend creating an environment file and specify the ``export PBM_MONGODB_URI=$PBM_MONGODB_URI`` statement within.
+
+As an example, let's configure to run backup snapshots on 23:30 every Sunday. 
+The steps are the following:
+
+1. Create an environment file. Let's name it :file:`pbm-cron`. 
+   The path for this file on Debian and Ubuntu is :file:`/etc/default/pbm-cron`. For Red Hat Enterprise Linux and CentOS, the path is :file:`/etc/sysconfig/pbm-cron`. 
+
+#. Specify the environment variable in :file:`pbm-cron`:
+   
+   .. code-block:: guess
+   
+      export PBM_MONGODB_URI="mongodb://pbmuser:secretpwd@localhost:27018?/replSetName=xxxx" 
+
+#. Grant access to the :file:`pbm-cron` file for the user that will execute the ``cron`` task.
+#. Create a ``crontab`` file. Let's name it ``pbm-backup``. 
+#. Specify the command in the file:
+   
+   .. code-block:: guess
+   
+      30 23 * * sun <user-to-execute-cron-task> . /etc/default/pbm-cron; /usr/bin/pbm backup 
+
+ Note the dot ``.`` before the environment file. It sources (includes) the environment file for the rest of the shell commands.
+
+#. Verify that backups are running in :file:`/var/log/cron` or :file:`/var/log/syslog` logs:
+   
+   .. code-block:: guess
+   
+      $ grep CRON /var/log/syslog
+
+.. admonition:: Backup storage cleanup
+
+   Previous backups are not automatically removed from the backup storage. 
+   You need to remove the oldest ones periodically to limit the amount of space used in the backup storage. 
+
+   We recommend using the :command:`pbm delete backup --older-than <timestamp>` command. You can configure a ``cron`` task to automate backup deletion by specifying the following command in the :file:`crontab` file:
+
+   .. code-block:: guess
+
+      /usr/bin/pbm delete-backup -f --older-than $(date -d '-1 month' +\%Y-\%m-\%d)
+
+   This command deletes backups that are older than 30 days. You can change the period by specifying a desired interval for the ``date`` function.
+
+Schedule backups with |PITR| running
+---------------------------------------------------
+
+It is convenient to automate making backups on a schedule using ``crond`` if you enabled :ref:`pitr`. 
+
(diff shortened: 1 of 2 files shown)


doc/source/schedule-backup.rst
```diff
@@ -0,0 +1,47 @@
...
+   Therefore, timely remove them to clean up the storage. 
+
+Schedule backups with |PITR|
+---------------------------------------------------
+
+ You can both enable incremental backups for :ref:`pitr` and run backup snapshots on a schedule using ``crond``. The steps are the following:
+
+ 1. Enable point-in-time recovery
+ 2. Make a fresh backup manually. It will serve as the starting point for incremental backups
+ 3. Configure ``crond`` to run backup snapshots on a schedule
```
akira-kurogane, line 43:
```suggestion
  Once you are running full backup snapshots on a schedule using ``crond`` it becomes suitable to enable :ref:`pitr`. The steps below can be done in any order, but PITR will only start running after at least one full backup has been made.

 * Enable point-in-time recovery
 * Make a fresh backup manually. It will serve as the starting point for incremental backups
 * Configure ``crond`` to run backup snapshots on a schedule
```

akira-kurogane (changes requested):
Hi Anastasia.

Thanks for generating all this new content. I have some edit requests. What do you think?

doc/source/schedule-backup.rst
```diff
@@ -0,0 +1,47 @@
...
+   .. code-block:: guess
+   
+      $ grep CRON /var/log/syslog
+
+.. note:: 
+
+   Previous backups are not automatically removed from the backup storage. 
+   Therefore, timely remove them to clean up the storage. 
+
+Schedule backups with |PITR|
```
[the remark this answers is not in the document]
nastena1606: I mean here "schedule backups with PITR running"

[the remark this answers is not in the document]
akira-kurogane: Ah, I'll amend my suggestion above.

akira-kurogane, line 36:
```suggestion
Schedule backups with |PITR| running
```

akira-kurogane (changes requested):

doc/source/schedule-backup.rst
```diff
@@ -0,0 +1,51 @@
...
+The steps are the following:
+
+1. Create a ``crontab`` file. Let's name it ``pbm-backup``. 
+2. Specify the command in the file:
+   
+   .. code-block:: guess
+   
+      30 23 * * sun . /etc/default/pbm-agent; /usr/bin/pbm backup 
+
+ where the :file:`/etc/default/pbm-agent` is the environment variable file path for Debian and Ubuntu and the dot ``.`` equals to bash builtin ``source``.
```
akira-kurogane, line 25:
```suggestion
 Please note the dot ``.`` before /etc/default/pbm-agent - it sources (includes) the file for the rest of the shell commands being run. This allows you to put the "export PBM_MONGODB_URI=mongodb://username:password@host1,.../" statement in a file and have that variable in the ``pbm`` program's environment rather than supplied as a command line argument, which is insecure. 
 
`/etc/default/pbm-agent` is the default location used for the environment variable for ``pbm-agent`` services. This is where the PBM_MONGODB_URI variable is usually set. You can use another path if you like to, however.
```

doc/source/schedule-backup.rst
```diff
@@ -0,0 +1,51 @@
...
+3. Verify that backups are running in :file:`/var/log/cron` or :file:`/var/log/syslog` logs:
+   
+   .. code-block:: guess
+   
+      $ grep CRON /var/log/syslog
+
+.. note:: 
+
+   Previous backups are not automatically removed from the backup storage. 
+   Therefore, timely remove them to clean up the storage.
```
akira-kurogane, line 36:
```suggestion
   Previous backups are not automatically removed from the backup storage. You need to remove the oldest ones periodically to limit the amount of space used in the backup storage. The recommend solution is to run ``pbm delete-backup --older-than YYYY-MM-DD`` using a crontab entry similar to the one for ``pbm backup`` shown above.
```

akira-kurogane (changes requested):
Hi Anastasia.

I've made some more suggestions. I think this is the last set.

doc/source/schedule-backup.rst
sandraromanchenko: IMHO, using /etc/default/pbm-agent (or /etc/sysconfig/pbm-agent file for RPM based systems) isn't correct due to several reasons:
1) since it's default location for the env variable for pbm-agent service, it shouldn't contain 'export' command, the proper syntax is VARIABLE=VALUE
2) without export command env variable won't be available for pbm process during cron command execution

I think there are several ways to do this:
1) if it's really required to use default env file, cron command should be like this
30 23 * * sun root . /etc/default/pbm-agent; PBM_MONGODB_URI=$PBM_MONGODB_URI /usr/bin/pbm backup

BTW, if to create cron task under /etc/cron.d is necessary to set user (used to execute command) and in this case root is required since /etc/default/pbm-agent (or /etc/sysconfig/pbm-agent) is owned by root

2) create a separate env file (let's say /etc/default/pbm-cron) with syntax like
export PBM_MONGODB_URI="....."
then cron command should look like
30 23 * * sun any_user_with_access_to_env_file . /etc/default/pbm-cron; /usr/bin/pbm backup

sandraromanchenko (changes requested):

doc/source/schedule-backup.rst
nastena1606: @akira-kurogane Which variant should we describe?

doc/source/schedule-backup.rst
sandraromanchenko: As discussed during standup, delete command should be like this 
`/usr/bin/pbm delete-backup -f --older-than $(date -d '-1 month' +\%Y-\%m-\%d)`

sandraromanchenko (changes requested):

doc/source/schedule-backup.rst
akira-kurogane: Hi @sandraromanchenko 

You're right - the /etc/default/pbm-agent file, consumed by systemd in it's service-launching procedure, is the "VAR=value" syntax. This clashes with the "export VAR=value" format it would have to be as a shell include file, for the crond command line. I was lax in that I didn't install pbm-agent properly and went from memory as to what was in the file. My apologies that I've sent both of you on this wild goose chase. Also I was wrong in thinking that su'ing as root was avoidable if using the /etc/default/pbm-agent file.

I like the (1) method shown above:
`30 23 * * sun root . /etc/default/pbm-agent; PBM_MONGODB_URI=$PBM_MONGODB_URI /usr/bin/pbm backup`

 which brings the "VAR=value" /etc/default/pbm-agent file's value into the environment of the pbm CLI process being launched. Without causing it to be in the command line arguments that other users can see if they run 'ps -ef' etc, which is the original security problem we're trying to overcome.

But there will be users who don't have the /etc/default/pbm-agent file on the server they run this pbm CLI crond command from and/or who don't want to run as root, and they should use (2). So we'll have to document both methods I think.

sandraromanchenko (approved):

akira-kurogane (approved):
Thanks so much!

PR #537, opened 2021-04-26
PBM-659 Note about backup incompatibility with PBM 1.4.1 and 1.5.0

nastena1606:
	modified:   source/upgrading.rst

### DIFF
--- a/doc/source/running.rst
+++ b/doc/source/running.rst
@@ -84,6 +84,13 @@ For |PBM| version 1.3.4 and earlier, run the |pbm-list| command and you will see
 Restoring a backup
 --------------------------------------------------------------------------------
 
+.. warning::
+
+   Backups made with |PBM| prior to v1.5.0 are incompatible for restore with |PBM| v1.5.0 and later. This is because processing of system collections ``Users`` and ``Roles`` has changed: in v1.5.0, ``Users`` and ``Roles`` are copied to temporary collection during backup and must be present in the backup during restore. In earlier versions of |PBM|, ``Users`` and ``Roles`` are copied to a temporary collection during restore. Therefore, restoring from these backups with |PBM| v1.5.0 isn't possible. 
+
+   The recommended approach is to make a fresh backup after :ref:`upgrading Percona Backup for MongoDB <pbm.upgrade>` to version 1.5.0.
+
+
 To restore a backup that you have made using |pbm-backup|, use the
 |pbm-restore| command supplying the time stamp of the backup that you intend to
 restore.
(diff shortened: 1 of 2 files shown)


doc/source/upgrading.rst
```diff
@@ -22,8 +22,9 @@ You can upgrade |PBM| to the **latest version** or to a **specific version**. Si
 .. rubric:: Important notes
  
 1. Backward compatibility between data backup and restore is supported for upgrades within one major version only (for example, from 1.1.x to 1.2.y). When you upgrade |PBM| over several major versions (for example, from 1.0.x to 1.2.y), we recommend to make a backup right after the upgrade.
-2. Starting from version 1.3.0, |pbm| packages are stored in the *pbm* repository and the *tools* repository for backward compatibility. To upgrade to the version lower than 1.3.0, enable the *tools* repository.
-3. Upgrade |PBM| on all nodes where it is installed.
+2. |PBM| v1.5.0 and later is incompatible with |PBM| v1.4.1 and earlier due to different processing of system collections ``Users`` and ``Roles`` during backup / restore operations. After the upgrade to |PBM| v1.5.0 and later, make sure to make a fresh backup.
+3. Starting from version 1.3.0, |pbm| packages are stored in the *pbm* repository and the *tools* repository for backward compatibility. To upgrade to the version lower than 1.3.0, enable the *tools* repository.
```
akira-kurogane, line 26: I think we should remove this advice regarding how to find the packages for 1.2.x or lower.

akira-kurogane (approved):
Hi Anastasia.

I've added a comment, I don't think we need to explain how to find the packages for < v1.3 nowadays.

Apart from that it all looks look good.

sandraromanchenko (approved):

doc/source/running.rst
```diff
@@ -84,6 +84,13 @@ For |PBM| version 1.3.4 and earlier, run the |pbm-list| command and you will see
 Restoring a backup
 --------------------------------------------------------------------------------
 
+.. warning::
+
+   Backups made with |PBM| v1.4.1 and earlier are incompatible for restore with |PBM| v1.5.0 and later. This is because processing of system collections ``Users`` and ``Roles`` has changed: in v1.5.0, ``Users`` and ``Roles`` are copied to temporary collection during backup and must be present in the backup during restore. In v1.4.1 and earlier, ``Users`` and ``Roles`` are copied to a temporary collection during restore. Therefore, restoring from these backups with |PBM| v1.5.0 isn't possible.
```
dAdAbird, line 89: Maybe it makes sense not to mention any specific version prior to v.1.5.0. I mean instead phrasing: 
```
Backups made with |PBM| prior to v1.5.0 are incompatible for restore with |PBM| v1.5.0 and later.
```
And so on.

Just in case we find some critical bug or gonna have some other reason to release the patch for v1.4 - v1.4.2. In that case, have to remember to change docs here etc.

Plus it more obvious - everything prior 1.5.0 isn't compatible.

dAdAbird: @akira-kurogane what do you think? ^^^

akira-kurogane: I didn't see a problem with saying 1.4.1 explicitly, given this is a patch for the imminent release of v1.5.0 and we don't have separate version branches, the chances of there being a 1.4.2 are now virtually zero.

Having said that, I do like the suggested edit more. It's shorter to read, simple.

PR #539, opened 2021-04-27
PBM-490: sync backup start time

dAdAbird:
Although we do synchronise oplog last write across agents in the sharded
cluster, the first write captured oplog is differ across the cluster. Hence
the backup start isn't synchronised. There is no way to start snapshot
(mongodump) at the same cluster time across the nodes. But we can do
that for oplog capturing.

This commit makes oplog start time (cluster time) to be the same across
all nodes. So oplog coverage now is consistent across the cluster not
only for the last write but for the first write as well.

https://jira.percona.com/browse/PBM-490


akira-kurogane (approved):
Arigato!

PR #554, opened 2021-06-07
PBM-685 Described additional install options:

nastena1606:
install from tarball
linked to DockerHub image.
Corrected directive syntax for running.rst
modified:   source/installation.rst

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -7,15 +7,15 @@ Installing |pbm|
 .. contents::
    :local:
 
-|percona| provides and supports installation packages for |pbm| in the *deb* and
-*rpm* formats that you can install by using ``apt`` or ``yum`` or other
-interfaces to your package management system.
+|percona| provides and supports |pbm| installation packages for Debian, Ubuntu, Red Hat Enterprise Linux and CentOS Linux distributions. Find detailed information about supported Linux distributions on the`Percona Software and Platform Lifecycle <https://www.percona.com/services/policies/percona-software-platform-lifecycle#mongodb>`_ page.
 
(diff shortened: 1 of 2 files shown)


doc/source/installation.rst
```diff
@@ -116,6 +131,39 @@ By running :program:`pbm version`, you can verify if |pbm| has been built correc
...
+You can `download installation packages <https://www.percona.com/downloads/percona-backup-mongodb/>`_ specific for your operating system from Percona website and install them using ``dpkg`` (Debian and Ubuntu) or ``rpm`` (Red Hat Enterprise Linux and CentOS). However, you must make sure that all dependencies are satisfied.
+
+Alternatively, you can download and install |pbm| from binary tarballs.
+
+Install from binary tarball
+---------------------------
+
+Find the link to the binary tarballs under the *Generic Linux* menu item on `Percona website <https://www.percona.com/downloads/percona-backup-mongodb/>`_.
+
+1. Fetch the correct binary tarball. For example, if you are running Debian 10 (“buster”):
```
sandraromanchenko, line 148: There is only one tarball, so I think maybe there is no need to set an example with buster (otherwise there is an impression that there are different tarballs for different OS)

sandraromanchenko (changes requested):

sandraromanchenko (approved):

doc/source/installation.rst
akira-kurogane: Yes, I recall that about half a year ago we changed from multiple linux tarballs to just one. So no surprise if we find older documentation that suggests there are a few different tarballs that are compiled linked with assumptions of a several different linux distributions, but I believe we create only one universal-use set of binaries for each product, which is built with more static linking and will run on any supported linux distro.

akira-kurogane (approved):

PR #559, opened 2021-07-01
PBM-403: hide user credential from ps output

dAdAbird:
When we extract creds from the argument the text became shorter while the mem size remains the same so we have to pad this difference with zeroes.
This leads to a bit strange output if shortened flag followed by any other flag - quite a few spaces between flags.

https://jira.percona.com/browse/PBM-403

### DIFF
--- a/cmd/pbm-agent/args.go
+++ b/cmd/pbm-agent/args.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"log"
 	"net/url"
 	"os"
@@ -22,7 +21,7 @@ func setarg(i int, as string) {
 	}
 }
 
-// hidecreds erases creds (user & pass) if there are any in mongo connection flags
+// hidecreds replaces creds (user & pass) with bunch of `x` if there are any in mongo connection flags
 func hidecreds() {
 	for k, v := range os.Args {
 		if strings.HasPrefix(v, mauthPrefix) {
@@ -36,9 +35,23 @@ func hidecreds() {
 				return
 			}
 
-			u.User = nil
-			fmt.Println(u.String())
+			var xuser, xpass []byte
+			xuser = make([]byte, len(u.User.Username()))
+			p, _ := u.User.Password()
+			xpass = make([]byte, len(p))
+
+			padx(xuser)
+			padx(xpass)
+
+			u.User = url.UserPassword(string(xuser), string(xpass))
+
 			setarg(k, mauthPrefix+u.String())
 		}
+	}
+}
+
+func padx(s []byte) {
+	for i := 0; i < len(s); i++ {
+		s[i] = 'x'
 	}
 }
--- a/cmd/pbm-agent/args.go
+++ b/cmd/pbm-agent/args.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+	"os"
+	"reflect"
+	"strings"
+	"unsafe"
+)
+
+const mauthPrefix = "--" + mongoConnFlag + "="
+
+func setarg(i int, as string) {
+	ptr := (*reflect.StringHeader)(unsafe.Pointer(&os.Args[i]))
+	arg := (*[1 << 30]byte)(unsafe.Pointer(ptr.Data))[:ptr.Len]
+
+	n := copy(arg, as)
+	for ; n < len(arg); n++ {
+		arg[n] = 0
+	}
+}
+
+// hidecreds erases creds (user & pass) if there are any in mongo connection flags
+func hidecreds() {
+	for k, v := range os.Args {
+		if strings.HasPrefix(v, mauthPrefix) {
+			str := strings.TrimPrefix(v, mauthPrefix)
+			u, err := url.Parse(str)
+			if err != nil {
+				log.Println(err)
+				return
+			}
+			if u.User == nil || u.User.String() == "" {
+				return
+			}
+
+			u.User = nil
+			fmt.Println(u.String())
+			setarg(k, mauthPrefix+u.String())
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


JNKPercona:
| Test name  | Status |
| Sharded cluster mongodb 3.6 | failed |
| Sharded cluster mongodb 4.0 | passed |
| Sharded cluster mongodb 4.2 | passed |
| Restore on new cluster mongodb 3.6 | passed |
| Restore on new cluster mongodb 4.2 | failed |
| Restore on new cluster mongodb 4.0 | passed |
| Non-sharded replicaset mongodb 3.6 | passed |
| Non-sharded replicaset mongodb 4.0 | passed |
| Single-node replicaset mongodb 4.0 | passed |
| Single-node replicaset mongodb 3.6 | passed |
| Non-sharded replicaset mongodb 4.2 | passed |
| Single-node replicaset mongodb 4.2 | passed |

akira-kurogane (commented):
Hi @Oneiroi . Could I ask your opinion about this patch?

The goal of this patch is so the pbm-agent will overwrite what is in its cmdline args that you see when running `ps`, or to put more technically what is seen when run cat /proc/<pid>/cmdline (at least for the linux case).

`pbm-agent --mongodb-uri=mongodb://akira:secret@hostabc.com:27917/`
-->

`pbm-agent --mongodb-uri=mongodb://hostabc.com:27917/            `

The method used has been to write into the raw string pointer from the os.Args list. I'm interpreting this as be the same as doing a sprintf(argv[x], "new shortened string") in a C program.

I thought this would only work by using the prctl() syscall with the PR_SET_MM action. But it seems that didn't succeed, but this does.

The user doesn't trigger the hidecreds() function that calls the setargs() function - it happens once. at startup.

JNKPercona:
| Test name  | Status |
| Sharded cluster mongodb 3.6 | passed |
| Sharded cluster mongodb 4.0 | passed |
| Sharded cluster mongodb 4.2 | passed |
| Restore on new cluster mongodb 4.0 | passed |
| Restore on new cluster mongodb 3.6 | passed |
| Non-sharded replicaset mongodb 3.6 | passed |
| Restore on new cluster mongodb 4.2 | failed |
| Non-sharded replicaset mongodb 4.0 | passed |
| Single-node replicaset mongodb 3.6 | passed |
| Non-sharded replicaset mongodb 4.2 | failed |
| Single-node replicaset mongodb 4.0 | passed |
| Single-node replicaset mongodb 4.2 | passed |

Oneiroi:
@akira-kurogane I've looked over the code, my notes are as follows:

1. Do we need to ensure that a SIGHUP / soft reload (read: re-read vars and restart process without exiting) still functions with this in place, assuming of course this app/binary is capable of servicing such a signal.
2. This should become a standalone lib such that it may be re-used for such purposes.
3. I do not at this moment in time, forsee issues in the use given the memory should still remain the same as this is producing memory replacement e.g. much more difficult (not impossible) to stage a payload into memory of the process.

Oneiroi:
@akira-kurogane I will note that the use of `unsafe.Pointer` may lead to issues, I am current researching this further, and will update, once I have a better understanding of the implications therein

Oneiroi:
From research, the general community consensus (AFAICT) is to use another method for the passing of the sensitive information, such as configuration file or other (which is supported in this application I belive), this approach whilst functional uses the `unsafe` package which is noted as providing unsafe operations: `package unsafe contains operations that step around the type safety of Go programs.` 

Now `unsafe` can be used, however to ensure that programmatic mistakes do not leak to unintended gaps/vulnerabilities requires several steps some of which are indicated in this blog: https://blog.gopheracademy.com/advent-2019/safe-use-of-unsafe-pointer/

Further research indicates this approach is also used to produce malware in golang, thus I would expect the use of this to start to (if not already doing so) cause EDR/NGAV to being to prevent operation due to this manipulation of memory.

Oneiroi:
Additional resources https://dev.to/jlauinger/exploitation-exercise-with-unsafe-pointer-in-go-information-leak-part-1-1kga this one notes a threat model for the use of `unsafe.Pointer` and does a fairly good job at describing the problems its used introduces.

Example PoCs where `unsafe.Pointer` is used including code-injection https://github.com/jlauinger/go-unsafepointer-poc 

Finally this covers where this is in use and what unsafe is for https://medium.com/a-journey-with-go/go-what-is-the-unsafe-package-d2443da36350 in detail

Oneiroi:
Taking code-injection as an example https://github.com/jlauinger/go-unsafepointer-poc/tree/master/code-injection the example Go code is here: https://github.com/jlauinger/go-unsafepointer-poc/blob/master/code-injection/main.go once built / executed, this can be exploited through Return Orientated Programming quoted the README.md file: 

>>  The exploit_rop.py file runs the program and supplies a carefully crafted input that overwrites the stack up to the point where the stored return address is located. It gets overwritten with a series of ROP gadgets, that is small fragments of assembly code that end with a ret instruction, thus chaining the gadgets is as simple as concatenating their addresses.

https://github.com/jlauinger/go-unsafepointer-poc/blob/master/code-injection/exploit_rop.py 

This seeks to prove that a buffer-overflow is introduced:

>>> A buffer of length 8 bytes is cast into a []byte slice with improper length configuration using the unsafe package and slice header representation from the reflect package. Then, using a bufio reader data is taken from the standard input and written to the slice, causing a buffer overflow. This is a size-constrained version of a classic gets()-based buffer overflow vulnerability.

Thus given `unsafe` allows for operations in which gloang does not apply regular checks and bounds, this _can_ lead to vulnerability , this in turn has many aspects to consider for risk acceptance if any.

As the examples provided require adjacent access to memory of the process in order to carry out the ROP exploitation, the risk could be stated as LOW, if however the use of the application is within a multi-tenant environment within which controls in place are lax or non existent, this would IMHO be HIGH, given potential of access to memory in order to carry out the ROP being likely.

akira-kurogane:
Ok, after much paranoid checking (sorry @dAdAbird!) I trust that we don't get a buffer overflow in setarg() even if the "as" string passed in might be abused.

akira-kurogane (approved):
LGTM

PR #566, opened 2021-07-19
PBM-617 Described the ability to restore from any backup

nastena1606:
modified:   source/pbm-commands.rst
	modified:   source/point-in-time-recovery.rst


doc/source/point-in-time-recovery.rst
```diff
@@ -76,16 +82,20 @@ Re-enable |PITR| to resume saving oplog slices:
...
+
+To restore from any backup snapshot, |PBM| requires continuous oplog. After the backup snapshot is made and |PITR| is re-enabled, it copies the oplog saved with the backup snapshot and creates oplog slices from the end time of the latest slice to the new starting point thus making the oplog continuous.
+
 
 .. rubric:: Delete a backup
 
-When you :ref:`delete a backup <pbm.backup.delete>`, all oplog slices that relate to this backup will be deleted too. For example, you delete a backup snapshot 2020-07-24T18:13:09 while there is another snapshot
-2020-08-05T04:27:55 created after it.  |pbm-agent| deletes only oplog slices that relate to 2020-07-24T18:13:09.
+When you :ref:`delete a backup <pbm.backup.delete>`, all oplog slices that relate to this backup will be deleted too. For example, you delete a backup snapshot 2020-07-24T18:13:09 while there is another snapshot 2020-08-05T04:27:55 created after it.  |pbm-agent| deletes only oplog slices that relate to 2020-07-24T18:13:09.
```
sandraromanchenko, line 92: This part is modified entirely within PBM-652, is this change still necessary?

doc/source/point-in-time-recovery.rst
```diff
@@ -56,9 +56,15 @@ Run |pbm-restore| and specify the timestamp from the valid range:
...
-   $ pbm restore --time="2020-07-14T14:27:04"
+   $ pbm restore --time="2020-12-14T14:27:04"
 
-Restoring to the point in time requires both a backup snapshot and oplog slices that can be replayed on top of it (the time ranges in PITR section of ``pbm list`` output must be later then the selected backup snapshot). 
+Restoring to the point in time requires both a backup snapshot and oplog slices that can be replayed on top of this backup. The time stamp you specify for the restore must meet the following conditions:
+
+* It must be from the time ranges in the PITR section of ``pbm list`` output
+* It must be later than the backup snapshot used for the restore. 
+
+To illustrate this behavior, let’s use the ``pbm list`` output from the previous example. For timestamp ``2020-07-14T14:27:04``,
```
sandraromanchenko, line 66: Pls correct the date 2020-07-14 to 2020-12-14

sandraromanchenko (changes requested):

doc/source/point-in-time-recovery.rst
nastena1606: I've reflected these changes in https://github.com/percona/percona-backup-mongodb/pull/568 so after the merge the doc will be consistent

doc/source/point-in-time-recovery.rst
nastena1606: Thanks. Fixed.

sandraromanchenko (approved):

doc/source/point-in-time-recovery.rst
```diff
@@ -56,9 +58,15 @@ Run |pbm-restore| and specify the timestamp from the valid range:
 
 .. code-block:: bash
 
-   $ pbm restore --time="2020-07-14T14:27:04"
+   $ pbm restore --time="2020-12-14T14:27:04"
+
+Restoring to the point in time requires both a backup snapshot and oplog slices that can be replayed on top of this backup. The time stamp you specify for the restore must meet the following conditions:
+
+* It must be from the time ranges in the PITR section of ``pbm list`` output
```
dAdAbird, line 65:
```suggestion
* It must be within the time ranges in the PITR section of ``pbm list`` output
```

doc/source/point-in-time-recovery.rst
```diff
@@ -56,9 +58,15 @@ Run |pbm-restore| and specify the timestamp from the valid range:
...
+* It must be later than the backup snapshot used for the restore.
```
dAdAbird, line 66: _If `--base-snapshot` is set. Otherwise, PBM will choose the base on it's own and the user doesn't have think about that - just follow the first condition_

doc/source/point-in-time-recovery.rst
dAdAbird: Still, I don't think we should have this paragraph anymore. As of v1.6 `pbm delete-backup` doesn't delete any pitr-chunks in any case.

dAdAbird (changes requested):

doc/source/point-in-time-recovery.rst
nastena1606: Won't this auto-selected snapshot be earlier in time than the specified PITR timestamp?

dAdAbird: It will. I just meant that if the user doesn't set `--base-snapshot` they shouldn't think about any snapshot. Any time that follows the first condition implicitly has some backup that finished before this time and pbm will choose that backup automatically. So the user doesn't have to worry about any snapshot in such a case.

Maybe
```
If `--base-snapshot` is used the timestamp must be later than the backup snapshot used for the restore.
```

nastena1606: Rewritten

dAdAbird (approved):

PR #603, opened 2021-10-15
Pbm 714 new backup pitr alignment  no txn

dAdAbird:
1. Wakeup sig from the backup to PITR now is the separate sig with the backup's opid. So PITR slicer knows what backup to wait for (to sync with) without the necessity of having a backup lock or meta beforehand.
2. Sending sig is a no more blocking operation so the system can progress (have elections etc).
3. Instead of a rather sophisticated scheme for the lock acquisition when the backup routine on each agent created an in-memory backup intent so PITR lock won’t be tried to reinstall. Then waited for 15+1 sec to be sure that the PITR routine has observed intent and only then swap locks. But this is anyway unreliable since everything happens in memory of every node  - no guarantees that all nodes received backup cmd and created intents - and relies on wall time  - no guarantees that all nodes received backup cmd the more or less same time and 15+1 is sufficient. With transactions everything would have been pretty straightforward - backup’s lock acquires consists of two ops in txn - delete pitr lock and install backup’s loc. But I do think that even without transactions this approach is better than the previous algorithm. Of course, there is a possibility that the lock would be discarded due to PITR squeeze its own lock in-between delete-install. But big chances there are would be (an)other agents to try to install the backup’s lock. Even if we fail - ok, discard backup, next time. This is a low probability event. 

Also:
1. Drop support of Mongo 3.6
2. Fix cluster versions in Jenkins.
3. Fix leader heartbeats during the restore. Sometimes followers used to omit the leader's heartbeats at the start and backed off with the error.

Fix https://jira.percona.com/browse/PBM-714

### DIFF
--- a/cli/restore.go
+++ b/cli/restore.go
@@ -138,6 +138,9 @@ func waitForRestoreStatus(ctx context.Context, cn *pbm.PBM, name string) error {
 		case <-tk.C:
 			fmt.Print(".")
 			meta, err = cn.GetRestoreMeta(name)
+			if errors.Is(err, pbm.ErrNotFound) {
+				continue
+			}
 			if err != nil {
 				return errors.Wrap(err, "get metadata")
 			}
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -49,6 +49,11 @@ func (b *Backup) Run(ctx context.Context, bcp pbm.BackupCmd, opid pbm.OPID, l *p
 }
 
 func (b *Backup) Init(bcp pbm.BackupCmd, opid pbm.OPID, balancer pbm.BalancerMode) error {
+	ts, err := b.cn.ClusterTime()
+	if err != nil {
+		return errors.Wrap(err, "read cluster time")
+	}
+
 	meta := &pbm.BackupMeta{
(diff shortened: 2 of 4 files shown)


JNKPercona:
| Test name  | Status |
| Restore on new cluster mongodb 4.2 | passed |
| Restore on new cluster mongodb 4.4 | passed |
| Restore on new cluster mongodb 4.0 | passed |
| Sharded cluster mongodb 4.0 | passed |
| Sharded cluster mongodb 4.2 | failed |
| Sharded cluster mongodb 4.4 | failed |
| Sharded cluster mongodb 5.0 | passed |
| Restore on new cluster mongodb 5.0 | passed |
| Non-sharded replicaset mongodb 4.0 | passed |
| Non-sharded replicaset mongodb 4.4 | passed |
| Single-node replicaset mongodb 4.0 | passed |
| Non-sharded replicaset mongodb 4.2 | passed |
| Single-node replicaset mongodb 4.2 | passed |
| Non-sharded replicaset mongodb 5.0 | passed |
| Single-node replicaset mongodb 4.4 | failed |
| Single-node replicaset mongodb 5.0 | failed |

PR #613, opened 2021-11-29
[PBM-750] Make max number of S3 upload parts configurable

dalbani:
https://jira.percona.com/projects/PBM/issues/PBM-750

### DIFF
--- a/pbm/storage/s3/s3.go
+++ b/pbm/storage/s3/s3.go
@@ -69,6 +69,9 @@ func (c *Conf) Cast() error {
 			}
 		}
 	}
+	if c.MaxUploadParts <= 0 {
+		c.MaxUploadParts = s3manager.MaxUploadParts
+	}
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pbm/storage/s3/s3.go
```diff
@@ -143,6 +144,10 @@ func (s *S3) Save(name string, data io.Reader, sizeb int) error {
 			}
 		}
 
+		maxUploadParts := s3manager.MaxUploadParts
+		if s.opts.MaxUploadParts > 0 {
+			maxUploadParts = s.opts.MaxUploadParts
+		}
```
dAdAbird, line 150: The same as with [the previous PR](https://github.com/percona/percona-backup-mongodb/pull/615) I'd move it to the Cast()

Cheers!

dAdAbird (changes requested):
Hey @dalbani 
Thanks for another contrib!

nastena1606 (approved):

pbm/storage/s3/s3.go
dalbani: I've move the code inside `Cast()`. Is that what you wanted?

dAdAbird (approved):

PR #619, opened 2021-12-15
[PBM-756] Support configurable compression level for backups

dalbani:
https://jira.percona.com/browse/PBM-756

### DIFF
--- a/cli/backup.go
+++ b/cli/backup.go
@@ -13,8 +13,9 @@ import (
 )
 
 type backupOpts struct {
-	name        string
-	compression string
+	name             string
+	compression      string
+	compressionLevel []int
 }
 
 type backupOut struct {
@@ -44,11 +45,17 @@ func runBackup(cn *pbm.PBM, b *backupOpts, outf outFormat) (fmt.Stringer, error)
 		return nil, errors.Wrap(err, "get remote-store")
 	}
 
+	var level *int
+	if len(b.compressionLevel) > 0 {
+		level = &b.compressionLevel[0]
+	}
+
 	err = cn.SendCmd(pbm.Cmd{
 		Cmd: pbm.CmdBackup,
 		Backup: pbm.BackupCmd{
-			Name:        b.name,
-			Compression: pbm.CompressionType(b.compression),
+			Name:             b.name,
+			Compression:      pbm.CompressionType(b.compression),
+			CompressionLevel: level,
 		},
 	})
 	if err != nil {
--- a/pbm/backup/dst.go
+++ b/pbm/backup/dst.go
@@ -70,13 +70,10 @@ func Compress(w io.Writer, compression pbm.CompressionType, level *int) (io.Writ
 			switch *level {
 			case 1:
 				writerOptions = append(writerOptions, s2.WriterUncompressed())
-				break
 			case 3:
 				writerOptions = append(writerOptions, s2.WriterBetterCompression())
-				break
 			case 4:
 				writerOptions = append(writerOptions, s2.WriterBestCompression())
-				break
 			}
 		}
 		return s2.NewWriter(w, writerOptions...), nil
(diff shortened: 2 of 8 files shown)


dalbani:
By the way, why isn't Jenkins triggered and the E2E tests not being run in case of a PR just like mine?

cli/backup.go
```diff
@@ -44,11 +45,17 @@ func runBackup(cn *pbm.PBM, b *backupOpts, outf outFormat) (fmt.Stringer, error)
 		return nil, errors.Wrap(err, "get remote-store")
 	}
 
+	var level *int
```
dalbani, line 48: If someone has a suggestion to avoid this dirty hack, I'm open to suggestion 😀 
But I couldn't find any other way with the Kingpin library to support an optional integer argument which has to be `nil` when not provided.

dAdAbird: I propose using `-1` as the default. kinping's flags have Default() functions so we're safe here. 
The only potential pitfall is the straightforward usage of [BackupCMD](https://github.com/percona/percona-backup-mongodb/pull/619/files#diff-71edcf47fb2c9f5a756a0b6aa78b3a5609368d7cf52fdcf64acf8660ab0eeec2R158) But we can add some comments there so a user would be aware of default bechaviour. 

BTW, I took a look and pgzip's DefaultCompression is -1 as well

pbm/backup/dst.go
```diff
@@ -23,29 +24,63 @@ type NopCloser struct {
...
 		if cc == 0 {
 			cc = 1
 		}
-		return s2.NewWriter(w, s2.WriterConcurrency(cc))
+		writerOptions := []s2.WriterOption{s2.WriterConcurrency(cc)}
+		if level != nil {
+			switch *level {
+			case 1:
+				writerOptions = append(writerOptions, s2.WriterUncompressed())
+				break
```
dAdAbird, line 73: no need in `break`. In Go control doesn't fall to the next case by default (https://go.dev/ref/spec#Switch_statements)

speedt/speedt.go
```diff
@@ -142,7 +142,7 @@ func Run(nodeCN *mongo.Client, stg storage.Storage, compression pbm.CompressionT
 
 	r := &Results{}
 	ts := time.Now()
-	size, err := backup.Upload(context.Background(), src, stg, compression, fileName, -1)
+	size, err := backup.Upload(context.Background(), src, stg, compression, nil, fileName, -1)
```
dAdAbird, line 145: It makes sense to add compression level to the speed-test as well

dAdAbird (changes requested):

dAdAbird:
> By the way, why isn't Jenkins triggered and the E2E tests not being run in case of a PR just like mine?
@dalbani 
Due to security reasons, we don't trigger Jenkins with the code from external contributors. We mulling thru options on tests triggered manually by maintainers etc. But this is a low priority task for the time being (despite it being a wee stone in the shoe)

pbm/backup/dst.go
dalbani: Thanks for the tip, I know only the basics of Go 👍

cli/backup.go
dalbani: There were 2 reasons why I didn't pick `-1` as equivalent of default instead of `nil`:
1. There could be compression methods that support/use `-1` as a valid, non-default value. Actually, I see that the Zstandard documentation mentions such levels: https://facebook.github.io/zstd/zstd_manual.html.
2. It doesn't play well with the `omitempty` functionality in Go, where a non specified integer gets `0` assigned.

As in:
```golang
type BackupCmd struct {
	Name             string          `bson:"name"`
	Compression      CompressionType `bson:"compression"`
	CompressionLevel int             `bson:"level,omitempty"`
}
```

On the other hand, having `CompressionLevel *int` as I proposed, means being able to differentiate between `0` as "no value passed" and `0` as "level 0 specifically".

dAdAbird (approved):

defbin (approved):
lgtm

PR #622, opened 2021-12-23
[PBM-764] Support Zstandard compression format

dalbani:
https://jira.percona.com/browse/PBM-764

### DIFF
--- a/pbm/backup/dst.go
+++ b/pbm/backup/dst.go
@@ -2,6 +2,7 @@ package backup
 
 import (
 	"compress/gzip"
+	"github.com/klauspost/compress/zstd"
 	"io"
 	"runtime"
 
@@ -77,6 +78,8 @@ func Compress(w io.Writer, compression pbm.CompressionType, level *int) (io.Writ
 			}
 		}
 		return s2.NewWriter(w, writerOptions...), nil
+	case pbm.CompressionTypeZstandard:
+		return zstd.NewWriter(w)
 	default:
 		return NopCloser{w}, nil
 	}
--- a/pbm/backup/dst.go
+++ b/pbm/backup/dst.go
@@ -79,7 +79,11 @@ func Compress(w io.Writer, compression pbm.CompressionType, level *int) (io.Writ
 		}
 		return s2.NewWriter(w, writerOptions...), nil
 	case pbm.CompressionTypeZstandard:
-		return zstd.NewWriter(w)
+		encLevel := zstd.SpeedDefault
+		if level != nil {
+			encLevel = zstd.EncoderLevelFromZstd(*level)
+		}
+		return zstd.NewWriter(w, zstd.WithEncoderLevel(encLevel))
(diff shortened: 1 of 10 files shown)


pbm/backup/dst.go
```diff
@@ -45,6 +46,9 @@ func Compress(w io.Writer, compression pbm.CompressionType) io.WriteCloser {
 			cc = 1
 		}
 		return s2.NewWriter(w, s2.WriterConcurrency(cc))
+	case pbm.CompressionTypeZstandard:
+		zw, _ := zstd.NewWriter(w)
```
dAdAbird, line 50:
```suggestion
		// an error is always nil with no options
		zw, _ := zstd.NewWriter(w)
```

dalbani: Thanks for your feedback, @dAdAbird.
Although I must say that it's not really clear to me what the original intent was in terms of error handling around `Compress()`.
I mean, if I look at where `Compress()` is called (e.g. <https://github.com/percona/percona-backup-mongodb/blob/main/pbm/backup/backup.go#L441>), I don't see the return value being checked for a `nil` value.
Shouldn't `Compress()` itself have a return signature of `(io.WriteCloser, error)` for proper error handling??

dAdAbird: Hi @dalbani !
`Compress()` doesn't return any errors since nothing inside it may trigger any. `zw, _ := zstd.NewWriter(w)` won't return errors since there unless it takes some options. I just wanted to express that in the comment to explain why the error isn't handled there. 
But, given that we have https://github.com/percona/percona-backup-mongodb/pull/619 which brings `Compress() (io.WriteCloser, error)` I'd rather wait for the 619 to be merged. And then add zstd along with the compression level. If this makes sense to you.

And can you also add it to the `pbm-speed-test`.

Thanks!

defbin (approved):
lgtm

pbm/backup/dst.go
dalbani: > But, given that we have #619 which brings Compress() (io.WriteCloser, error) I'd rather wait for the 619 to be merged. And then add zstd along with the compression level. If this makes sense to you.

#619 has been reviewed, so if you merge it first, I can then update the code here to match the new error handling implemented there.

defbin: #619 has been merged

dalbani:
I have rebased the code on `main`.

defbin (changes requested):
You do not provide compression level support for the zstd. Could you please add it for consistency

dalbani:
Doh! Who's this guy who added this compression level thing 🤪
Will do.

defbin (approved):
thanks

dalbani:
I've just added `zstd` to the list of compression types accepted by commands.

defbin (approved):

dAdAbird (approved):

PR #629, opened 2022-01-26
[PBM-777] Expand/fix config related API

dalbani:
https://jira.percona.com/browse/PBM-777

### DIFF
--- a/pbm/config.go
+++ b/pbm/config.go
@@ -294,6 +294,9 @@ func (p *PBM) GetConfigVar(key string) (interface{}, error) {
 		return nil, errors.Wrap(err, "get from db")
 	}
 	v, err := bts.LookupErr(strings.Split(key, ".")...)
+	if err != nil {
+		return nil, errors.Wrap(err, "lookup in document")
+	}
 	switch v.Type {
 	case bson.TypeBoolean:
 		return v.Boolean(), nil
--- a/pbm/config.go
+++ b/pbm/config.go
@@ -227,8 +227,8 @@ func (p *PBM) SetConfigVar(key, val string) error {
 	// just check if config was set
 	_, err := p.GetConfig()
 	if err != nil {
-		if errors.Cause(err) == mongo.ErrNoDocuments {
-			return errors.New("config doesn't set")
+		if errors.Is(err, mongo.ErrNoDocuments) {
+			return errors.New("config is not set")
 		}
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pbm/config.go
```diff
@@ -269,6 +269,28 @@ func (p *PBM) SetConfigVar(key, val string) error {
...
 }
 
+func (p *PBM) DeleteConfigVar(key string) error {
+	if !ValidateConfigKey(key) {
+		return errors.New("invalid config key")
+	}
+
+	_, err := p.GetConfig()
+	if err != nil {
+		if errors.Cause(err) == mongo.ErrNoDocuments {
```
defbin, line 279: please use `errors.Is()` function. `errors.Cause()` returns a direct underlying error in a chain. `mongo.ErrNoDocuments` could be not direct but wrapped by another somewhere in `GetConfig()`

defbin (changes requested):

pbm/config.go
dalbani: Actually I copied this piece of code from `GetConfigVar` above.
But I'm ready to change both of them for consistency.

defbin: you are welcome to change both :)

pbm/config.go
```diff
@@ -304,7 +329,7 @@ func (p *PBM) GetConfigVar(key string) (interface{}, error) {
 	case bson.TypeDouble:
 		return v.Double(), nil
 	case bson.TypeString:
-		return v.String(), nil
+		return v.StringValue(), nil
```
dalbani, line 332: I have changed this instruction from `v.String()` to `v.StringValue()` to avoid getting the quoted value back.
But there are probably side effects that I didn't see.

pbm/config.go
dalbani: Done.

pbm/config.go
defbin: @dAdAbird do you have a reason to keep using .String()?

defbin (approved):
lgtm

defbin (approved):

pbm/config.go
dalbani: @dAdAbird  
Did you have the time to have a look at it, by any chance?

dalbani: Ping 😄

dAdAbird (approved):

PR #643, opened 2022-03-29
PBM-801 Config file example

nastena1606:


packaging/conf/pbm-conf-reference.yml
```diff
@@ -0,0 +1,112 @@
...
+
+#====================Point-in-Time Recovery Configuration==================
+
+#pitr:
+#  enabled: false
+
+#Adjust the size op oplog chunks (in minutes).
+#  oplogSpanMin: 10 
+
+#Set a compression method and level
```
defbin, line 95: pls ensure space character between `#` and the next letter character for consistency with the whole example

defbin (approved):

PR #644, opened 2022-03-30
PBM-808, PBM-807: CLI - waiting for the logical restore operations

dAdAbird:
- Added "wait" to the PITR restore and Oplog relay.
- Check for the operation heartbeats during the wait. That will prevent indefinite wait if agents are gone. All nodes in non-sharded env - is the only agent that runs a restore.

### DIFF
--- a/cli/oplog.go
+++ b/cli/oplog.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"time"
 
@@ -8,6 +9,28 @@ import (
 
 	"github.com/percona/percona-backup-mongodb/pbm"
 )
+
+type replayOptions struct {
+	start string
+	end   string
+	wait  bool
+}
+
+type oplogReplayResult struct {
+	Name string `json:"name"`
+	done bool
+	err  string
+}
+
+func (r oplogReplayResult) String() string {
+	if r.done {
+		return "\nOplog replay successfully finished!\n"
+	}
+	if r.err != "" {
+		return "\n Error: " + r.err
+	}
+	return fmt.Sprintf("Oplog replay %q has started", r.Name)
+}
 
 func replayOplog(cn *pbm.PBM, o replayOptions, outf outFormat) (fmt.Stringer, error) {
 	startTS, err := parseTS(o.start)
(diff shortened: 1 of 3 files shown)


JNKPercona:
| Test name  | Logical | Physical |
| ------------- | ------------- | ------------- |
| New cluster psmdb 4.4 | passed | - |
| New cluster psmdb 4.2 | passed | - |
| Sharded psmdb 4.2 | failed | passed |
| Sharded psmdb 4.4 | failed | passed |
| Non-sharded psmdb 4.2 | passed | failed |
| Single-node psmdb 4.2 | passed | passed |
| Single-node psmdb 4.4 | passed | failed |
| Non-sharded psmdb 4.4 | passed | passed |
| Sharded psmdb 5.0 | failed | failed |
| New cluster psmdb 5.0 | passed | - |
| Non-sharded psmdb 5.0 | failed | failed |
| Single-node psmdb 5.0 | failed | failed |

cli/oplog.go
```diff
@@ -9,6 +10,28 @@ import (
...
+}
+
+func (r oplogReplayResult) String() string {
+	if r.done {
+		return "\nOplog replay successfully finished!\n"
+	}
+	if r.err != "" {
+		return "\n Error: " + r.err
+	}
+	return fmt.Sprintf("Oplog replay %q has started", r.Name)
```
defbin, line 32: `fmt.Stringer` is used to get a string representation of a value that can be used anywhere.
here I see you are using it to format output specifically (~de~coupled) to CLI. I'd use a separate interface/method for CLI output. fallback to `String()` if there is no implementation for CLI output

defbin, line 32: `fmt.Stringer` is used to get a string representation of a value that can be used anywhere.
here I see you are using it to format output specifically (~de~coupled) to CLI. I'd use a separate interface/method for CLI output. fallback to `String()` if there is no implementation for CLI output

defbin (approved):

cli/oplog.go
dAdAbird: Yes, maybe, need to think about it. But it is used in CLI where the object has to be in plain text, hence string. Maybe we have to have a separate interface which includes Stringer, json.Marshaler etc

PR #691, opened 2022-08-18
[PBM-922]: Add support for SSE-C server-side encryption for S3 endpoints

taxilian:
I have completed a partial test of this -- I am able to initialize pbm, list status etc, start a backup. I'm still running that backup so we'll see how it does.

I can confirm that the files written to the S3 repository are encrypted with SSE-C, however.

Generate a SSE-C AES256 key with:

```bash
openssl rand -base64 32
```

Example config file for Backblaze B2:

```yaml
storage:
  type: s3
  s3:
    region: us-west-002
    endpointUrl: https://s3.us-west-002.backblazeb2.com
    bucket: pbm-test-bucket-name
    prefix: data/pbm/backup
    credentials:
      access-key-id: <redacted>
      secret-access-key: <redacted>
    serverSideEncryption:
      sseCustomerAlgorithm: AES256
      sseCustomerKey: "eBYL81+sAigzkanckAeKQ0kitmVAPwN2DfbItdeMlR8="
```

I'll finish my tests but I wanted to get this pushed up for other eyes to look and see if I've missed anything. Unlike using KMS the `sseCustomerAlgorithm` and `sseCustomerKey` options need to be set on any S3 api call which reads or writes files. Deleting does not seem to need it =] I think I got them all, though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ You can install Percona Backup for MongoDB in the following ways:
 
 Find the installation instructions in the [official documentation](https://www.percona.com/doc/percona-backup-mongodb/installation.html)
 
-Alternaively, you can [run Percona Backup for MongoDB as a Docker container](https://hub.docker.com/r/percona/percona-backup-mongodb).
+Alternatively, you can [run Percona Backup for MongoDB as a Docker container](https://hub.docker.com/r/percona/percona-backup-mongodb).
 
 ## Submit Bug Report / Feature Request
 
--- a/pbm/config.go
+++ b/pbm/config.go
@@ -45,6 +45,9 @@ func (c Config) String() string {
 	if c.Storage.S3.Credentials.Vault.Token != "" {
 		c.Storage.S3.Credentials.Vault.Token = "***"
 	}
+	if c.Storage.S3.ServerSideEncryption.SseCustomerKey != "" {
+		c.Storage.S3.ServerSideEncryption.SseCustomerKey = "***"
+	}
 	if c.Storage.Azure.Credentials.Key != "" {
 		c.Storage.Azure.Credentials.Key = "***"
 	}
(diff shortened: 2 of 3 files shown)


taxilian:
I have now tested a full backup, PITR snapshots, and a full restore to a new collection. I have also updated it so `pbm config` will not output the encryption key and added experimental support for both KMS and SSE-C encryption for minio endpoints -- worst case on that is if it doesn't work it effectively changes nothing, but I don't have a minio target to test it with.

As far as I can tell this PR should be ready for merge, but please let me know if there is anything else you want me to look at with it.

dAdAbird:
Thanks for the contribution! LGTM, but we'd have to run tests on our side before merging.

taxilian:
One issue I did hit but don't see any mitigation for other things:

When I switched to use this I was unable to do a `pbm config` until I had written a new config first -- From the error message (something about accessing `Nil`, IIRC) I assume this is because it didn't find anything for those fields.

Is there something I need to do to set a default value for those? How is that usually handled? I don't expect it will be a difficult fix and the workaround is just to set the config again (I just used a file) but it does seem like something which should be fixed.

taxilian:
Is there anything more you need for this to be usable? I saw an update in jira earmarking this for a 2.0 release, but no word on time line or anything. I need it to be able to use with our production cluster at GradeCam, so trying to figure out the best path forward =]

sandraromanchenko (approved):

dAdAbird:
@taxilian thanks a lot for the contribution!
I made a commit to fix `pbm config` [panic](https://github.com/percona/percona-backup-mongodb/pull/691#issuecomment-1219807611)

taxilian:
Awesome, I'll have to go look at what you did -- I couldn't find anywhere else that something like that was implemented so wasn't sure how to address it.

taxilian:
Ahh -- I see what I missed now. Good info, I'll watch out for that in the future. Sorry I basically have no real experience with golang, just occasionally little things like this where I just copied what I saw ;-) I look forward to seeing this in a real release!

PR #742, opened 2022-11-22
PBM-998: fix json decoding error

neohe:
The contents of the mongod log file is not a single JSON object but a series of unrelated JSON objects.
Decode multiple JSON values from the source one-by-one.


dAdAbird (approved):

dAdAbird:
Thanks for the contribution!

PR #750, opened 2022-12-04
Fix CheckRestore function in e2e-tests

ziollek:
Add missing return statement if restore error has been found. Without a return statement restore is not recognized as a failure until it times out (by default 25 minutes).

Signed-off-by: ziollek <e.prace@gmail.com>


defbin (approved):

dAdAbird:
Thanks!

dAdAbird (approved):

PR #769, opened 2023-01-10
Fix some more typos

deejay1:
Some more typos fixes in user facing messages.
Also one fix for a variable name (`s/predule/prelude/`)


defbin (approved):
thanks

PR #736, opened 2022-10-19
feat: add aws IRSA credentials support

pastequo:
Hello,

In the continuity of this PR (https://github.com/percona/percona-backup-mongodb/pull/592), this PR adds support for IRSA

PR #772, opened 2023-01-16
feat: Support Access Key free GCS authent

pastequo:
Hello,

In the continuity of the PR https://github.com/percona/percona-backup-mongodb/pull/736, this PR allows to connect to GCS without Access Key ID / Secret Access Key. It uses kubernetes & GCP service account and workload identity.

- To write actual backup files, minio can't be used (it doesn't support this kind of authN). If I understood correctly, they are not going to implement it (https://github.com/minio/minio/issues/14264). So I used directly GCS SDK that is surprisingly simple. More tests would be welcome to ensure it works fine even with large amount of data

- To perform other operation, I changed a little bit the s3 client following this article: https://github.com/salrashid123/aws_s3_gcs_sdk
It allows to put on the fly the right Authorization header for GCP, based on something a little bit hacky (but unfortunately necessary).
Another approach could be to reimplement all methods for the GCS provider and not benefit from the s3 compatible API

Let me know if the proposal here seems correct or not to you


pastequo:
go.mod had to be completed. Since the vendor is commit-ed in the repo, the PR seems to contain a lot of changes, in fact 2 files are created/modified

pastequo:
Any feedback ?

miecio45:
Hello everyone. This would be really nice feature, any feedback would be great.

capri1989:
Hi, 

sorry for the very late reply here. Something that "scares" me a little, is the need of so many more libraries, which cause more testing and adds way more complexity and risk. Do we see another way of achieving this with less additions?

Kai

PR #803, opened 2023-03-15
chore: Rename module for v2 version

pastequo:
Hey,

Since v2.0.0 go module naming doesn't follow go recommendation (module should be named "github.com/percona/percona-backup-mongodb/v2") and can't be imported easily. For example in the mongodb operator: https://github.com/percona/percona-server-mongodb-operator/blob/v1.14.0/go.mod#L17

Go authors recommended for a while to create a v2/ folder and maintain several version of the code (https://go.dev/blog/v2-go-modules) but doesn't seem to reach a vast adhesion in the community. I personally - as many open source projects - prefer the approach to rename the module and the imports but keep the folder organization. Let me know what you think about it.

Sorry if it has already been discussed, I haven't found anything about it

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -11,10 +11,10 @@ import (
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 
-	"github.com/percona/percona-backup-mongodb/pbm"
-	"github.com/percona/percona-backup-mongodb/pbm/log"
-	"github.com/percona/percona-backup-mongodb/pbm/storage"
-	"github.com/percona/percona-backup-mongodb/version"
+	"github.com/percona/percona-backup-mongodb/v2/pbm"
+	"github.com/percona/percona-backup-mongodb/v2/pbm/log"
+	"github.com/percona/percona-backup-mongodb/v2/pbm/storage"
+	"github.com/percona/percona-backup-mongodb/v2/version"
 )
 
 type Agent struct {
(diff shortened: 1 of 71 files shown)


defbin:
`v2` is good for public API. PBM does not provide stable Golang public API. It's designed as binaries. The only public API is `pbm` CLI. Any function in the repo can be changed at any time.
So before switching to a `v2`-like module, we must establish Golang public API that will follow "semver".

Secondly, I do not see a reason to introduce `v2` within the ongoing major release. This is better to do at least the next major `v3` release.

The recommendation was designed to follow the semantic version or, at least, be API-compatible within a major version. As for now, we cannot guarantee we won't introduce breaking changes in the public function, type, etc. Doing it now will restrict us.

Basically, `"github.com/percona/percona-backup-mongodb/*"` is semantically private. 

also see [PBM-780](https://jira.percona.com/browse/PBM-780)

pastequo:
I see

Private packages are meant to be in `internal/` but as mentioned in the Jira "The operator uses internal functions which are not considered public.", which prevents you from doing it.

As a feedback, I found confusing the combination of non-`internal/` packages + the tagging strategy for binaries colliding with the golang way of publishing package version =!=> the packages can be used.  It made me think that those packages are public. Any other tagging convention (to avoid that [1.13](https://github.com/percona/percona-server-mongodb-operator/blob/v1.13.0/go.mod#L15) vs [1.14](https://github.com/percona/percona-server-mongodb-operator/blob/v1.14.0/go.mod#L17)), or a message in the readme, would have prevent me from losing my time. But that might be just me 😉 

I let you close the PR

defbin:
added message to readme about it (https://github.com/percona/percona-backup-mongodb/pull/804)

PR #828, opened 2023-04-18
PBM-1099 Updated pbm config reference

nastena1606:
Added AWS token, mongod binary path, concurrent download options

### DIFF
--- a/packaging/conf/pbm-conf-reference.yml
+++ b/packaging/conf/pbm-conf-reference.yml
@@ -1,66 +1,66 @@
 ######################## PBM Configuration ############################
 
-# This file is a full configuration example documenting all 
-# options in comments. 
-#
-# You can find the full configuration reference here:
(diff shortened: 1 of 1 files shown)


defbin (approved):

packaging/conf/pbm-conf-reference.yml
```diff
@@ -121,3 +122,14 @@
 #restore:
 #  batchSize: 500
 #  numInsertionWorkers: 10
+
+# Adjust concurrent download of data chunks from storage for physical restore.
```
dAdAbird, line 126:
```suggestion
## Adjust concurrent download of data chunks from storage for physical restore.
```

nastena1606: Updated

dAdAbird (approved):

PR #855, opened 2023-07-09
[PBM-1143] Add support for WebDAV backups

dalbani:
https://jira.percona.com/browse/PBM-1143

### DIFF
--- a/pbm/storage/webdav/webdav.go
+++ b/pbm/storage/webdav/webdav.go
@@ -1,0 +1,125 @@
+package webdav
+
+import (
+	"github.com/percona/percona-backup-mongodb/pbm/log"
+	"github.com/percona/percona-backup-mongodb/pbm/storage"
+	"github.com/pkg/errors"
+	"github.com/studio-b12/gowebdav"
+	"io"
+	"os"
+	"path"
+	"strings"
+)
+
+type Conf struct {
+	ServerURL   string      `bson:"serverUrl,omitempty" json:"serverUrl" yaml:"serverUrl,omitempty"`
+	Prefix      string      `bson:"prefix,omitempty" json:"prefix,omitempty" yaml:"prefix,omitempty"`
+	Credentials Credentials `bson:"credentials" json:"-" yaml:"credentials"`
+}
+
+type Credentials struct {
+	Username string `bson:"username" json:"username,omitempty" yaml:"username,omitempty"`
+	Password string `bson:"password" json:"password,omitempty" yaml:"password,omitempty"`
+}
+
+type WebDAV struct {
+	opts   Conf
+	log    *log.Event
+	client *gowebdav.Client
+}
+
+func New(opts Conf, l *log.Event) (*WebDAV, error) {
+	w := &WebDAV{
+		opts: opts,
+		log:  l,
+	}
+
+	w.client = gowebdav.NewClient(opts.ServerURL, opts.Credentials.Username, opts.Credentials.Password)
+	err := w.client.Connect()
+	if err != nil {
+		return nil, errors.Wrap(err, "connect to WebDAV server")
+	}
+
+	return w, nil
+}
+
+func (*WebDAV) Type() storage.Type {
+	return storage.WebDAV
+}
+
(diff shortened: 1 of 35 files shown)


pbm/storage/webdav/webdav.go
```diff
@@ -0,0 +1,125 @@
...
+	"io"
+	"os"
+	"path"
+	"strings"
+)
+
+type Conf struct {
+	ServerURL   string      `bson:"serverUrl,omitempty" json:"serverUrl" yaml:"serverUrl,omitempty"`
+	Prefix      string      `bson:"prefix,omitempty" json:"prefix,omitempty" yaml:"prefix,omitempty"`
+	Credentials Credentials `bson:"credentials" json:"-" yaml:"credentials"`
```
dalbani, line 17: I've copied this over from `s3.go` and `azure.go`.
But I'm wondering, is there a reason why it's set to `-` only for `json`?

defbin: Not sure why. I guess it to not be shown on debug output. The `bson:` is required to be stored in MongoDB.

dalbani: 👍

dalbani:
What's missing to this PR is maybe an integration test? Maybe it's not obvious to me how to do that.

defbin:
try to add it here https://github.com/percona/percona-backup-mongodb/blob/86d552880b047573b1be93de493186b266d60f75/e2e-tests/cmd/pbm-test/run.go#L28-L31

dalbani:
OK, then I need to fire up a WebDAV server, as I can rely on a external service like S3 or Azure.

defbin:
to run the test:
`env MONGODB_VERSION=6.0 TESTS_BCP_TYPE=logical e2e-tests/run-all`

on Apple Silicon you can try:
`env MONGODB_VERSION=6.0.6-arm64 TESTS_BCP_TYPE= logical e2e-tests/run-all`

dalbani:
Although it's all green, I am not sure if what I pushed actually run the E2E test for WebDAV 🤨

Speaking of E2E setup, I'm wondering if this choice of `path-ignores` is correct: https://github.com/percona/percona-backup-mongodb/blob/4ed23327208d9de4f01384e25911ceb14bd03686/.github/workflows/ci.yml#L22

In the sense that:
> When all the path names match patterns in paths-ignore, the workflow will not run. If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

For example, I only committed to `e2e-tests/pkg/tests/sharded/cluster.go` in https://github.com/percona/percona-backup-mongodb/pull/855/commits/96b095ea952616a3e6efd2f0ddb205b88e260b47, and, yet, CI ran.

capri1989:
Hi @dalbani,

just for me to better understand the use-case of WebDAV in 2023. In which scenario and where do you actually need WebDAV? Can you elaborate a little more on your needs?

Thanks,
Kai

dalbani:
Well, even though it's 2023, WebDAV is very much "alive".
I can name for example two storage providers which offer WebDAV-backed repositories — and no S3 compatible endpoint:
- [Hetzner Storage Box](https://docs.hetzner.com/robot/storage-box/)
- Microsoft OneDrive — yep, it's technically possible to connect via WebDAV, you can look it up 😃 

And there are other (open source) implementations of WebDAV servers available for self-hosted deployments.

capri1989:
Hey,

we've investigated and surveyed our current customer base, and there isn't genuine interest in WebDAV. Given that we need to maintain the code for a very long time, given support and SLAs, we don't consider adding it to the project.

So I'll close this PR, so you don't wait any longer on a decision that might never happen. Nevertheless thanks a lot for the contribution.

dalbani:
Thanks for your feedback.
In the end, I've come to the conclusion that *logical* backups are not suited to my use case.
Due to the size of my database, the generated backup files were like 200+ GB (!), even with the best compression.
So I need to have a look at physical backups, is my understanding.

PR #859, opened 2023-07-11
[PBM-1148] Simplify Dockerfile.k8s

dalbani:
https://jira.percona.com/browse/PBM-1148


docker/Dockerfile.k8s
```diff
@@ -38,14 +32,6 @@ RUN set -ex; \
     percona-release setup psmdb-50; \
```
dalbani, line 32: And what about upgrading to `psmdb-60`?
Given that:
1. PMB may have to interact with PSMDB v6 instances
2. `percona-server-mongodb-shell` v6 is backwards compatible with older versions (I suppose)

defbin:
this file is not more maintained

PR #862, opened 2023-07-18
Fixed broken links to old documentation references

capri1989:
There were four links in the README that didn't exist anymore and only showed a 404 or wrong pages.

I updated the links accordingly so that you end up on the right locations as well as showing the architecture picture again in the README:


nastena1606 (approved):
Thanks, LGTM

defbin (approved):

PR #900, opened 2023-12-15
[PBM-1223] backup oplog from the oldest active transaction timestamp

defbin:


olexandr-havryliak (approved):

pbm/oplog/oplog.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/percona/percona-backup-mongodb/pbm/errors"
+)
+
+var errNoTransaction = errors.New("no transaction found")
+
+func GetOplogStartTime(ctx context.Context, m *mongo.Client) (primitive.Timestamp, error) {
+	ts, err := findTransactionStartTime(ctx, m)
+	if errors.Is(err, errNoTransaction) {
+		ts, err = findLastOplogTS(ctx, m)
```
boris-ilijic, line 20: Maybe it should be better to fall back to old criteria if we cannot find such trx:
```
topo.OpTimeFromNodeInfo(inf, true)
```

boris-ilijic (approved):
Do not have full context but code LGTM.

PR #923, opened 2024-03-05
PBM_tests. Add delay for PITR, Return OL8 as base image

sandraromanchenko:


e2e-tests/pkg/tests/sharded/trx.go
```diff
@@ -361,6 +361,7 @@ func (c *Cluster) checkTrxCollection(ctx context.Context, col string, bcp Backup
 	}
 	c.flushRouterConfig(ctx)
 
+	time.Sleep(time.Second * 60)
```
defbin, line 364: 😬

defbin (approved):
lgtm

e2e-tests/pkg/tests/sharded/trx.go
sandraromanchenko: Not cool, but it's the easiest way to avoid sporadic test failures as from time to time PITR doesn't catch up to the desired restore point(

PR #927, opened 2024-03-14
[PBM-1252] keep list of physical files in storage

defbin:
https://jira.percona.com/browse/PBM-1252

### DIFF
--- a/cmd/pbm/main.go
+++ b/cmd/pbm/main.go
@@ -132,7 +132,7 @@ func main() {
 	backupCmd.Flag("wait", "Wait for the backup to finish").
 		Short('w').
 		BoolVar(&backupOptions.wait)
-	backupCmd.Flag("list-files", "Wait for the backup to finish").
+	backupCmd.Flag("list-files", "Shows the list of files per node (external backup only)").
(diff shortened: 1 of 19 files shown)


cmd/pbm/main.go
```diff
@@ -132,7 +132,7 @@ func main() {
 	backupCmd.Flag("wait", "Wait for the backup to finish").
 		Short('w').
 		BoolVar(&backupOptions.wait)
-	backupCmd.Flag("list-files", "Wait for the backup to finish").
+	backupCmd.Flag("list-files", "Shows the list of files per node (external backup only)").
```
boris-ilijic, line 135: fyi: This is part of another commit which is already in `dev`

pbm/version/version.go
```diff
@@ -15,7 +16,7 @@ import (
 )
 
 // current PBM version
-const version = "2.4.0"
+const version = "2.4.1-dev"
```
boris-ilijic, line 19: Why do not we use just: `2.4.1` ?

boris-ilijic (approved):

PR #932, opened 2024-03-26
pbm cli: fix 'follow logs' option to respect output format (json, jso…

michaello1:
Command 'pbm logs --follow --out=json' ignores output formatting (similarly for --out=json-pretty) and prints logs in plain text format. This change fixes formatting so all 3 formats are supported.

### DIFF
--- a/cmd/pbm/main.go
+++ b/cmd/pbm/main.go
@@ -431,7 +431,7 @@ func main() {
 	case cleanupCmd.FullCommand():
 		out, err = doCleanup(ctx, conn, pbm, &cleanupOpts)
 	case logsCmd.FullCommand():
-		out, err = runLogs(ctx, conn, &logs)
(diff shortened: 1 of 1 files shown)


defbin:
Hi. Thank you for contributing.

defbin:
Can you explain your use case?The `-f/--follow` flag is designed for plain text output. For structural logs, you can use `pbm logs --out json --tail 100`.

michaello1:
We are using it in a systemd service with redirected stdout to collect logs (as they are coming) to a central logging system. Mongod logs are in JSON format and PBM cli supports JSON output as well, so we wanted to keep it the same, even the structure is a bit different. It is not clearly documented that -f is designed for plain text only, so we fixed / modified it for our purpose.

defbin (approved):

boris-ilijic (approved):

boris-ilijic:
@michaello1, thank you for your contribution.

PR #940, opened 2024-05-07
PBM-1311: PBM fails to do PITR slicing when RS doesn't have a majority

boris-ilijic:
Fix for https://perconadev.atlassian.net/browse/PBM-1311

When doing PITR slicing PBM relies on `majorityOpTime` to define slicing intervals:
```
db.isMaster()
{
  ...
  lastWrite: {
    opTime: { ts: Timestamp({ t: 1715079500, i: 1 }), t: Long('1') },
    lastWriteDate: ISODate('2024-05-07T10:58:20.000Z'),
    majorityOpTime: { ts: Timestamp({ t: 1715079500, i: 1 }), t: Long('1') },
    majorityWriteDate: ISODate('2024-05-07T10:58:20.000Z')
  },
 ...
```
When cluster doesn't have majority `majorityOpTime` is stale,  so it contains last value when majority was present.

This PR fixes that using approach: when write concern is below majority, PBM uses `opTime` instead of `majorityOpTime`.

### DIFF
--- a/pbm/topo/cluster.go
+++ b/pbm/topo/cluster.go
@@ -7,6 +7,7 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/writeconcern"
 
 	"github.com/percona/percona-backup-mongodb/pbm/connect"
 	"github.com/percona/percona-backup-mongodb/pbm/errors"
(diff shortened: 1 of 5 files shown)


pbm/topo/cluster.go
```diff
@@ -66,6 +67,33 @@ func OpTimeFromNodeInfo(inf *NodeInfo, majority bool) (primitive.Timestamp, erro
...
+		return true, nil
+	}
+
+	s, err := GetReplsetStatus(ctx, m)
+	if err != nil {
+		return true, errors.Wrap(err, "get replset status")
+	}
+
+	return w >= s.WriteMajorityCount,
+		nil
```
defbin, line 94:
```suggestion
	return w >= s.WriteMajorityCount, nil
```

defbin (approved):

PR #949, opened 2024-05-28
Use RemoteCredsProvider from aws default stack

atorkhov:
This change allows to retrieve remote credentials not only from EC2, but also from ECS and EKS roles.
See https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html for details about EKS Pod Identities.


defbin:
https://jira.percona.com/browse/PBM-1351

atorkhov:
> hey @atorkhov can you please provide you use case? How do you want to authenticate your S3 connection for backups? 
> Why existing methods - AWS role or Access/Secret keys are not sufficient?

We authenticate using AWS role with help of EKS Pod Identity - https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html

defbin (approved):

boris-ilijic (approved):

PR #952, opened 2024-06-19
PBM tests. Avoid checking storage unless it's s3

sandraromanchenko:


boris-ilijic:
@sandraromanchenko In general I am fine with this change, but why we are skipping Azure and FS ?

sandraromanchenko:
@boris-ilijic it's just temporary hack, I won't merge it, so please ignore it for now)

boris-ilijic:
Yes, in the meantime I deleted my question, because I noticed that pr is in draft. Didn't noticed it before, sorry for confusion ...

defbin (approved):
I am okay with this fix. However, filesystem and Azure Storage checks for leftover files need to be implemented. At present, I'd do this separately to not block this fix.

boris-ilijic (approved):
I guess that we have the Jira ticket for the missing feature.

PR #954, opened 2024-06-25
PBM-1329 Write to temporary file name and sync before renaming

DanielOliverRJ:
Whilst trying to understand historic cases where we have seen BSON deserialisation errors, I realised it was possible for filesystem writes to be incomplete, but be unable to determine the fact.  In particular, once an oplog file exists on the filesystem, it is assumed to be complete even though the write may have been interrupted.

This change performs all filesystem writes to a temporary file in the same directory and does a rename once the sync has returned.

In the event of a failure, calling Close() on the same file handle does not cause an error because the deferred call discards the error.

I do not believe non-filesystem storage are affected by this problem, as the providers have other checks in place to ensure the upload completed.

### DIFF
--- a/pbm/storage/fs/fs.go
+++ b/pbm/storage/fs/fs.go
@@ -79,10 +79,7 @@ func (*FS) Type() storage.Type {
 	return storage.Filesystem
 }
 
-func (fs *FS) Save(name string, data io.Reader, _ int64) error {
-	filepath := path.Join(fs.root, name+".tmp")
-	finalpath := path.Join(fs.root, name)
-
+func WriteSync(filepath string, data io.Reader) error {
 	err := os.MkdirAll(path.Dir(filepath), os.ModeDir|0o755)
 	if err != nil {
 		return errors.Wrapf(err, "create path %s", path.Dir(filepath))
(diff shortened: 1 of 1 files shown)


DanielOliverRJ:
@defbin Unfortunately I am unable to review/sign the CLA as the website does not appear to be responding correctly.  After signing in, I am re-presented with the sign-in screen.

defbin:
hi @DanielOliverRJ. thank you for your contributions. Could you please try to review/sign the CLA again so we can proceed?

DanielOliverRJ:
Sorry, I had to get the CLA text approved before I could sign.  This is now completed.

defbin:
Hi @DanielOliverRJ 

Thank you for your contribution.

I see one problem here: https://github.com/percona/percona-backup-mongodb/blob/fd2c57a45acbbbe28f2f48f10c6407468a387ec6/pbm/storage/fs/fs.go#L70-L101

Please ensure that the temporary file is deleted in case of an error before the function exists. Thus, the caller code may remain unaware of the implementation details specific to file system storage.

DanielOliverRJ:
I've re-factored slightly to move write to a common function and attempt to remove the temporary file on any error.

defbin (approved):
tested by debugging

boris-ilijic (approved):

PR #957, opened 2024-06-27
PBM-1343: Release the lock when failed to read the storage config

t-yrka:
When the `pbm-agent` (during the pitr procedure) fails on reading the storage config it may get locked out of starting the slicer (which results in missing pitr data for a given replicaset in sharded cluster).

Each node is blocked at: https://github.com/percona/percona-backup-mongodb/blob/v2.5.0/cmd/pbm-agent/restore.go#L183-L185

Release the lock before exiting the function to allow another node to continue the slicing.


sandraromanchenko (approved):

boris-ilijic:
@t-yrka thank you for your contribution

PR #958, opened 2024-06-27
PBM-1332 Fix: failed to delete the history backup data

Cookiery:
Ticket: https://perconadev.atlassian.net/browse/PBM-1332

I've encountered the same problem as the ticket above, I debugged it and found that it's because the DeleteObject method doesn't support deleting directory Objects.

BTW, thanks for your contributions, it's a great tool to use!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ Before submitting code contributions, we ask you to complete the following prere
 
 ### 1. Sign the CLA
 
-Before you can contribute, we kindly ask you to sign our [Contributor License Agreement](https://cla-assistant.percona.com/&lt;linktoCLA>) (CLA). You can do this in on click using your GitHub account.
+Before you can contribute, we kindly ask you to sign our [Contributor License Agreement](https://cla-assistant.percona.com/) (CLA). You can do this in on click using your GitHub account.
 
 **Note**:  You can sign it later, when submitting your first pull request. The CLA assistant validates the PR and asks you to sign the CLA to proceed.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ Before submitting code contributions, we ask you to complete the following prere
 
 ### 1. Sign the CLA
 
-Before you can contribute, we kindly ask you to sign our [Contributor License Agreement](https://cla-assistant.percona.com/) (CLA). You can do this in on click using your GitHub account.
+Before you can contribute, we kindly ask you to sign our [Contributor License Agreement](https://cla-assistant.percona.com/percona/percona-backup-mongodb) (CLA). You can do this in on click using your GitHub account.
 
 **Note**:  You can sign it later, when submitting your first pull request. The CLA assistant validates the PR and asks you to sign the CLA to proceed.
 
--- a/pbm/storage/s3/download.go
+++ b/pbm/storage/s3/download.go
@@ -44,7 +44,7 @@ import (
 // would be allocated as some of the span slots may remain unused.
 
 const (
-	downloadChuckSizeDefault = 8 << 20
+	downloadChunkSizeDefault = 8 << 20
 	downloadRetries          = 10
 
 	ccSpanDefault = 32 << 20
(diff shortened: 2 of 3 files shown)


Cookiery:
Also corrected a few typo errors

Cookiery:
Hi @defbin , could you take a look my pull request, thanks!

defbin:
hi @Cookiery. thank you for your contribution.

I tested your branch with MinIO server, and it does not delete the `filelist.pbm` file. does it work for you?

Cookiery:
hi @defbin . thanks for your feedback. 
Sorry, I was so focused on deleting directories that I forgot about the .pbm.json file, I'll fix that.

defbin:
We don't need this logic. Just specify the file explicitly, like in https://github.com/percona/percona-backup-mongodb/pull/965/files, and the backup dir will be deleted at the end.

Please make it simple and just enough to fix the issue. You are free to copy changes from the alternative PR (I'll close it in favor of yours)

Cookiery:
I think it would be better to support deleting directories and files directly in the `Delete`function?

Cookiery:
The problem is that specifying a file doesn't delete the directory on S3, we need to list all files to delete it.

Cookiery:
Hi @defbin , I can delete the .pbm.json file with this function.
```
func main() {
	endpoint := ""
	accessKey := ""
	secretKey := ""

	sess, err := session.NewSession(&aws.Config{
		Region:           aws.String("us-east-1"),
		Endpoint:         aws.String(endpoint),
		S3ForcePathStyle: aws.Bool(true),
		Credentials:      credentials.NewStaticCredentials(accessKey, secretKey, ""),
	})

	if err != nil {
		fmt.Println("Failed to create session:", err)
		return
	}

	svc := s3.New(sess)

	bucket := "mongodb-backup"
	prefix := ""
	name := "2024-06-27T18:59:51Z.pbm.json"

	allObjects := []*s3.ObjectIdentifier{}

	err = svc.ListObjectsV2Pages(&s3.ListObjectsV2Input{
		Bucket: aws.String(bucket),
		Prefix: aws.String(path.Join(prefix, name)),
	}, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
		for _, obj := range page.Contents {
			allObjects = append(allObjects, &s3.ObjectIdentifier{
				Key: aws.String(*obj.Key),
			})
		}
		return !lastPage
	})

	if err != nil {
		fmt.Println("Failed to list objects:", err)
		return
	}

	if len(allObjects) == 0 {
		fmt.Println("No objects found to delete")
		return
	}

	// Batch delete
	var objectsToDelete []*s3.ObjectIdentifier
	for _, object := range allObjects {
		objectsToDelete = append(objectsToDelete, &s3.ObjectIdentifier{
			Key: object.Key,
		})
	}

	_, err = svc.DeleteObjects(&s3.DeleteObjectsInput{
		Bucket: aws.String(bucket),
		Delete: &s3.Delete{
			Objects: objectsToDelete,
			Quiet:   aws.Bool(true),
		},
	})

	if err != nil {
		fmt.Println("Failed to delete objects:", err)
		return
	}

	// Confirm delete
	for _, object := range objectsToDelete {
		err = svc.WaitUntilObjectNotExists(&s3.HeadObjectInput{
			Bucket: aws.String(bucket),
			Key:    object.Key,
		})

		if err != nil {
			fmt.Printf("Error occurred while waiting for object %s to be deleted: %v\n", aws.StringValue(object.Key), err)
		} else {
			fmt.Printf("Object %s deleted successfully\n", aws.StringValue(object.Key))
		}
	}
}
```

defbin:
did you tried the https://github.com/percona/percona-backup-mongodb/pull/965 ?

defbin:
I mean, did you try to delete a backup with that code changes only? What problem did you meet?

Cookiery:
I have the same problem as this ticket, that is, when deleting the backup file, none of the S3 directory files are deleted. 

https://perconadev.atlassian.net/browse/PBM-1332

image

Cookiery:
It's not a complicated function, just list all the objects and batch delete them.

defbin:
Please test the `defbin/PBM-1332` branch. It must resolve the issue. I tested the branch with MinIO and AWS S3 - and it works as expected.

PBM deletes files individually for a backup. The original issue is that one file is missing (which was added in a recent release).

Your PR adds new/extra logic (and network calls per file), which does not resolve the issue (I haven't found a practical reason to accept your changes). Did you analyze the impact of your changes? The Delete() method is called in many places. Have you reviewed each one?

Cookiery:
Thanks @defbin , you can close this PR.

PR #972, opened 2024-07-31
drop support MongoDB v4.4

defbin:


sandraromanchenko (dismissed):

sandraromanchenko (approved):

radoslawszulgo (approved):
In general OK. However, I don't know why do you update it to 5.0 and not to 7.0 - to avoid doing this again at the end of the year when 5.0 becomes EOL.

boris-ilijic (approved):

PR #1017, opened 2024-09-18
PBM-1114: check backup files before mark it as done

defbin:
backup status can be set to done in db meta before the meta is written to storage. if the write fails, no meta will be available on storage. storage resync will delete backup meta from db and won't see it on storage.

### DIFF
--- a/pbm/backup/query.go
+++ b/pbm/backup/query.go
@@ -81,15 +81,34 @@ func getBackupMeta(ctx context.Context, conn connect.Client, clause bson.D) (*Ba
 }
 
 func ChangeBackupStateOPID(conn connect.Client, opid string, s defs.Status, msg string) error {
-	return changeBackupState(context.Background(), conn, bson.D{{"opid", opid}}, s, msg)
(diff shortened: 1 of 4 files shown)


pbm/backup/query.go
```diff
@@ -81,15 +81,34 @@ func getBackupMeta(ctx context.Context, conn connect.Client, clause bson.D) (*Ba
...
-func changeBackupState(ctx context.Context, conn connect.Client, clause bson.D, s defs.Status, msg string) error {
-	ts := time.Now().UTC().Unix()
+func ChangeBackupStateWithUnix(
+	conn connect.Client,
+	bcpName string,
+	s defs.Status,
+	unix int64,
+	msg string,
+) error {
+	return changeBackupState(context.TODO(),
```
boris-ilijic, line 100: Just expand the method with context.Context and propagate the same `ctx` down.

boris-ilijic (dismissed):
I am fine with PR but please provide requested correction.

PR #1020, opened 2024-09-23
PBM-1397: improve backup files check

defbin:


pbm/util/errgroup.go
```diff
@@ -0,0 +1,44 @@
...
+	"runtime"
+	"sync"
+)
+
+type errorGroup struct {
+	errs []error
+	mu   sync.Mutex
+
+	wg  sync.WaitGroup
+	sem chan struct{}
```
boris-ilijic, line 13: This is nice semaphore implementation:
https://pkg.go.dev/golang.org/x/sync@v0.8.0/semaphore

PR #1021, opened 2024-09-23
Add support for Workload Identity Authentication

terryyanko:
Related to: https://github.com/percona/roadmap/issues/26

For workloads running on GCE, fetch the workload identity from the metadata server to allow for service account access to GCP services.


pbm/storage/s3/s3.go
```diff
@@ -617,3 +639,24 @@ func awsLogger(l log.LogEvent) aws.Logger {
 		l.Debug(msg, xs...)
 	})
 }
+
+func onGCE() bool {
+	client := http.Client{
+		Timeout: 100 * time.Millisecond,
+	}
+
+	req, err := http.NewRequest("GET", "http://169.254.169.254/computeMetadata/v1", nil)
```
egegunes, line 648: if we add the trailing slash, client doesn't need to follow redirect

```suggestion
	req, err := http.NewRequest("GET", "http://169.254.169.254/computeMetadata/v1/", nil)
```

pbm/storage/s3/s3.go
```diff
@@ -550,6 +558,20 @@ func (s *S3) session() (*session.Session, error) {
...
+	// If using GCE, attempt to retrieve access token from metadata server
+	if onGCE() {
+		tokenSource := google.ComputeTokenSource(s.opts.ServiceAccount, "")
+		token, err := tokenSource.Token()
+		if err != nil {
+			return nil, errors.Wrap(err, "get GCP token")
+		}
+		providers = append(providers, &credentials.StaticProvider{Value: credentials.Value{
+			AccessKeyID:     "GCP_OAUTH_TOKEN",
+			SecretAccessKey: "GCP_OATH_TOKEN",
```
egegunes, line 570: is `GCP_OATH_TOKEN` a typo?

terryyanko: Sorry for the delay, no not a typo, the fields are mandatory but they're just set to placeholder values here since they're not used for GCS.

pbm/storage/s3/s3.go
```diff
@@ -617,3 +639,24 @@ func awsLogger(l log.LogEvent) aws.Logger {
 		l.Debug(msg, xs...)
 	})
 }
+
+func onGCE() bool {
+	client := http.Client{
+		Timeout: 100 * time.Millisecond,
+	}
+
+	req, err := http.NewRequest("GET", "http://169.254.169.254/computeMetadata/v1/", nil)
```
radoslawszulgo, line 648: shouldn't "http://metadata.google.internal" be used instead of the hardcoded IP?

PR #1035, opened 2024-10-10
[PBM-1405] dump v2

defbin:


pbm/archive/conv.go
```diff
@@ -0,0 +1,133 @@
...
+	"strings"
+
+	mtArchive "github.com/mongodb/mongo-tools/common/archive"
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+
+	"github.com/percona/percona-backup-mongodb/pbm/storage"
+)
+
+func GenerateV1FromV2(stg storage.Storage, bcp, rs string) error {
```
boris-ilijic, line 18: Do we have the Jira ticket to align backup and restore about the metadata format? So currently we use V2->V1 conversion. In future we should make backup to export directly to V1, or we should make restore to work with V2 format directly.

pbm/archive/conv.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "open file: %q", metaV2Filename)
+	}
+	defer rdr.Close()
+
+	data, err := io.ReadAll(rdr)
+	if err != nil {
+		return errors.Wrap(err, "read")
+	}
+	rdr.Close()
```
boris-ilijic, line 30:
```suggestion
	err = rdr.Close()
	if err != nil {
		return errors.Wrap(err, "close")
	}
```
It'll be cleaner to add error handling here.

boris-ilijic (commented):
This looks good to me, please see few comments.

pbm/archive/backup.go
```diff
@@ -0,0 +1,351 @@
...
+
+func (bcp *backupImpl) writeMeta(collections []*NamespaceV2) error {
+	meta := &ArchiveMetaV2{
+		Version:       version.Current().Version,
+		ServerVersion: bcp.serverVersion,
+		FCV:           bcp.fcv,
+		Namespaces:    collections,
+	}
+
+	data, err := bson.Marshal(meta)
```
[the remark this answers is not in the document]
defbin: BSON is a better format for the data (direct support for binary (ObjectID), int64 (CRC, size), and compact storage size). JSON is text-only format. the meta file is for PBM only. not for end user. IMO, for testing, using bsondump is not a big deal (it is needed rarely)

pbm/archive/conv.go
defbin: v1 or v2 - does not matter. converting to v1 was done as quick solution. it is pbm format. when PBM creates archive for mongorestore, it uses data from the v1 but not format directly. PBM can makes the archive from v2. just some changes are needed in another place.

PR #1041, opened 2024-10-30
PBM-1414 Set Number of Parallel Collections to 1 by default

radoslawszulgo:


defbin:
maybe here https://github.com/percona/percona-backup-mongodb/blob/dev/pbm/snapshot/restore.go#L20

radoslawszulgo:
> maybe here https://github.com/percona/percona-backup-mongodb/blob/dev/pbm/snapshot/restore.go#L20

I'm not sure. I anyway need to declare this variable. I could do it with 0, and then always overwrite with what comes from the configuration. Would it be better?

defbin:
I meant, the request is about changing `--numInsertionWorkersPerCollection` to 1 (instead of 10). But your PR changes `--numParallelCollections`

boris-ilijic:
In addition by ticket requirement, the same parameter should be exposed as an option of `restore` command (e.g. `--num-insertion-workers`) which would override currently only available config parameter: `restore.numInsertionWorkers`
https://docs.percona.com/percona-backup-mongodb/reference/restore-options.html#restorenuminsertionworkers

cmd/pbm/main.go
```diff
@@ -244,6 +244,10 @@ func main() {
 		StringVar(&restore.pitrBase)
 	restoreCmd.Flag("num-parallel-collections", "Number of parallel collections").
 		Int32Var(&restore.numParallelColls)
+	restoreCmd.Flag("num-insertion-workers-per-collection",
+		"Specifies the number of insertion workers to run concurrently per collection. For large imports, "+
+			"increasing the number of insertion workers may increase the speed of the import.").
+		Int32Var(&restore.numInsertionWorkers)
```
defbin, line 250:
```suggestion
		Default("1").
		Int32Var(&restore.numInsertionWorkers)
```

set default value. in `pbm help restore` it will show `--num-insertion-workers-per-collection=1`

pbm/restore/logical.go
```diff
@@ -79,13 +80,13 @@ type oplogRange struct {
 type restoreUsersAndRolesOption bool
 
 // New creates a new restore object
-func New(
-	leadConn connect.Client,
+func New(leadConn connect.Client,
```
defbin, line 83:
```suggestion
func New(
	leadConn connect.Client,
```
keep consistency in style

pbm/restore/logical.go
```diff
@@ -1330,7 +1331,11 @@ func (r *Restore) applyOplog(ctx context.Context, ranges []oplogRange, options *
 }
 
 func (r *Restore) snapshot(input io.Reader, cloneNS snapshot.CloneNS, excludeRouterCollections bool) error {
-	rf, err := snapshot.NewRestore(r.brief.URI, r.cfg, cloneNS, r.numParallelColls, excludeRouterCollections)
+	rf, err := snapshot.NewRestore(r.brief.URI,
```
defbin, line 1334:
```suggestion
	rf, err := snapshot.NewRestore(
	r.brief.URI,
```
optionally but good for style consistency: keep each argument on a separate line so it is easy to see all of them. except for that `ctx`

boris-ilijic (changes requested):
In general LGTM, but please see some comments.

PR #1054, opened 2024-11-28
PBM-1444: Cloning for create/drop ops

boris-ilijic:
PR for https://perconadev.atlassian.net/browse/PBM-1444

This PR expands cloning collection feature with logic for processing `create` and `drop` operations from the oplog.
It uses the same approach as selective restore has.
In previous PR #1046 those 2 commands were just ignored.


veceraj (approved):
reviewed from go perspective

PR #1070, opened 2024-12-13
Bump all minor dependencies

boris-ilijic:


boris-ilijic:
This one is stale, new one will be created.

PR #1077, opened 2025-01-07
PBM-1466 - Remove kingpin dependency from pbm cli

veceraj:
PR for https://perconadev.atlassian.net/browse/PBM-1466

The PR replaces the use of kingpin by Cobra and Viper.


boris-ilijic (commented):
This PR is ready to be merged, but there are few suggestions that I'd prefer to improve.

Also we should remove completely `kingpin` dependency from `go.mod`, so let's do it here or in new PR (better).

veceraj:
Thanks, I like all the comments. Will update shortly.

PR #1083, opened 2025-01-22
Pbm-1474 Swap AWS s3 v1 -> v2 SDK API

veceraj:
... tbd

### DIFF
--- a/pbm/storage/s3/download.go
+++ b/pbm/storage/s3/download.go
@@ -2,19 +2,20 @@ package s3
 
 import (
 	"container/heap"
+	"context"
 	"crypto/md5"
 	"encoding/base64"
 	"fmt"
 	"io"
-	"net/http"
 	"path"
 	"runtime"
+	"strings"
 	"sync/atomic"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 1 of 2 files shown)


boris-ilijic (commented):
LGTM in general, please do the cleanup related to GCS handling. 
Beside that I am submitting some other suggestions from first part of the review, let's incorporate all and then I'll finalize review.

pbm/storage/s3/download.go
```diff
@@ -207,13 +209,13 @@ func (s *S3) newPartReader(fname string, fsize int64, chunkSize int) *partReader
...
-		getSess: func() (*s3.S3, error) {
-			sess, err := s.s3session()
+		getSess: func() (*s3.Client, error) {
+			cli, err := s.s3client()
 			if err != nil {
 				return nil, err
 			}
-			sess.Client.Config.HTTPClient.Timeout = time.Second * 60
-			return sess, nil
+			// sess.Client.Config.HTTPClient.Timeout = time.Second * 60
```
boris-ilijic, line 217: We'll need to specify timeout in within the context of each ops I guess.

boris-ilijic (approved):

boris-ilijic:
Hi @Pionerd, we are aware of the issue and we'll soon release fix for that. Please see the details here:
https://perconadev.atlassian.net/browse/PBM-1631

Solution is to upgrade to v2.12.0 which will be released approx. 2 weeks from now, or do downgrade to v2.9.1.

veceraj:
Hi @Pionerd, thanks for reaching out. I will think about this and look at UpCloud (I don't have experience with it). One possibility comes to my mind and that is to use the s3 compatible minio SDK. We are using Minio for backwards compatibility of GCS with hmac keys. I wonder if UpCloud could use the GCS type storage config.

PR #1096, opened 2025-02-19
PBM-1484 GCP SDK

veceraj:
PR for https://perconadev.atlassian.net/browse/PBM-1484

Migration from AWS SDK v1 to v2 introduced unexpected issues with Google Cloud Storage.
This feature implements the Google Cloud Storage package. 

### Credentials
- Currently supports Application Default Credentials - https://cloud.google.com/docs/authentication/application-default-credentials and Service account - https://cloud.google.com/iam/docs/service-account-overview
- HMAC credentials are not currently supported (handled separately in [PBM-1498](https://perconadev.atlassian.net/browse/PBM-1498))

Example storage config:

```yaml
storage:
  type: gcs
  gcs:
    bucket: pbm-bcp
    prefix: pbm/test
    credentials:
      projectId: psychic-mason-44...
      privateKey: -----BEGIN PRIVATE KEY-----\n ...
```
### ChunkSize
storage.gcs.chunkSize (optional) specified in bytes (https://pkg.go.dev/cloud.google.com/go/storage#Writer)
- defaults to nil - the GCS default is 16MB
- it is possible to specify 0 
  - From GCS docs: "If ChunkSize is set to zero, chunking will be disabled and the object will be uploaded in a single request without the use of a buffer."
- (Note: s3 has uploadPartSize, there is no equivalent to maxUploadParts)

[PBM-1498]: https://perconadev.atlassian.net/browse/PBM-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pbm/storage/gcs/gcs.go
```diff
@@ -0,0 +1,219 @@
...
+		bucketHandle: client.Bucket(opts.Bucket),
+		log:          l,
+	}, nil
+}
+
+func (*GCS) Type() storage.Type {
+	return storage.GCS
+}
+
+func (g *GCS) Save(name string, data io.Reader, size int64) error {
```
boris-ilijic, line 94: We need to have retrying logic here.

pbm/storage/gcs/gcs.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	if err := w.Close(); err != nil {
+		return errors.Wrap(err, "writer close")
+	}
+
+	return nil
+}
+
+func (g *GCS) SourceReader(name string) (io.ReadCloser, error) {
```
boris-ilijic, line 114: For physical restore and S3 we have integrated this logic with `Download` (concurrently downloading file from storage). Previously that logic was applied to GCS, so we do not have it anymore. We do not need to that within this PR imo, but we need to have story for that.

boris-ilijic (commented):
I am fine with this PR, it's simple and clean. Still we need to do some more improvements, please see the comments.

pbm/storage/gcs/gcs.go
```diff
@@ -0,0 +1,219 @@
...
+			opts.Credentials.PrivateKey,
+			opts.Credentials.ProjectID,
+			opts.Credentials.ProjectID,
+		)
+
+		credentials := option.WithCredentialsJSON([]byte(credStr))
+
+		client, err = gcs.NewClient(ctx, credentials)
+	} else {
+		client, err = gcs.NewClient(ctx)
```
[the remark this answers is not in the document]
veceraj: If no credentials are provided, the SDK check for the "application default credentials" env variable. 
When even that is not present, it ends with an error

pbm/storage/gcs/gcs.go
boris-ilijic: https://perconadev.atlassian.net/browse/PBM-1506

boris-ilijic (approved):

PR #1106, opened 2025-03-18
PBM-1522 Check full db name instead of a prefix

t-yrka:
Make sure to not skip other databases than 'admin'.

It aims to solve following issues:
1. When restoring a namespace with problematic database name (for example `administration.*`) it makes the database filter empty and ALL the users are restored (includes users from other databases).
2. When restoring multiple namespaces, including the problematic one (for example `administration.*,somedb.*`) it skips the`administation` database in filters and as a result the user related to this database is NOT restored (the other ones are).


boris-ilijic (approved):

veceraj (approved):

boris-ilijic:
Thanks for the contribution.

PR #1109, opened 2025-03-19
PBM-1494 Added condition if S3 EntypointURL is empty

dbazhenov:
Trying to fix a bug with S3
https://perconadev.atlassian.net/browse/PBM-1494

I think the error happens when the EndpointURL is empty. I added a check if it is empty. 
Since I can't test all variants, I made the check right after the path definition to leave the old logic unchanged.


boris-ilijic (approved):
Thanks for the contribution @dbazhenov !

PR #1115, opened 2025-03-26
Improve logging of physical restore

igroene:
The log messages are a bit cryptic. Hopefully this makes the flow of operations more clear.


igroene:
ok will keep it in mind for the future. Let me close this and redo

PR #1123, opened 2025-04-17
PBM-1498 HMAC support for GCS

veceraj:
This PR adds native HMAC authentication support for Google Cloud Storage (GCS) by introducing a dual-client approach. Previously, after migrating to the native GCS library (cloud.google.com/go/storage), HMAC keys stopped working because the library uses the JSON API, which does not support HMAC authentication (available only via the XML API). To restore HMAC compatibility, a new lightweight client was implemented using the MinIO SDK, which signs requests properly against GCS's S3-compatible XML API.

When users configure GCS storage, the system now automatically selects the appropriate client based on the credentials: service account JSON credentials will use the native GCS client (resumable uploads), while HMAC credentials will use the MinIO client (chunked/multipart uploads). Upload logic across S3, GCS (JSON), and GCS (HMAC) backends was unified with a new part-size computation helper, ensuring consistent multipart behavior under large files and service limits. Functionally, backup and restore workflows now behave the same for both credential types, requiring minimal or no migration effort for users upgrading from earlier PBM versions that accessed GCS via S3.


pbm/storage/gcs/hmac_client.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/percona/percona-backup-mongodb/pbm/storage"
+)
+
+type hmacClient struct {
+	client *minio.Client
+	opts   *Config
+	log    log.LogEvent
+}
+
+func newHmacClient(opts *Config, l log.LogEvent) (*hmacClient, error) {
```
boris-ilijic, line 24: According to convention should be `newHMACClient`

boris-ilijic (dismissed):

PR #1144, opened 2025-06-09
PBM-1544 Add new column showing the backup status update

veceraj:
WIP:
I we also want to display the printStatus when using `-o json`, the determination of the printStatus needs to happen earlier

### DIFF
--- a/cmd/pbm/status.go
+++ b/cmd/pbm/status.go
@@ -470,35 +470,22 @@ func (s storageStat) String() string {
 		return a.RestoreTS > b.RestoreTS
 	})
 
-	const (
-		statusSuccess = "success"
-		statusFailed  = "failed"
-		statusOngoing = "ongoing"
-	)
-
-	var printStatus string
-
 	for i := range s.Snapshot {
 		ss := &s.Snapshot[i]
 		var status string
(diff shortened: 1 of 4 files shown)


cmd/pbm/status.go
```diff
@@ -629,6 +617,7 @@ func getStorageStat(
 				snpsht.Err = errors.New(errStr)
 				snpsht.ErrString = errStr
 				snpsht.Status = defs.StatusError
+				snpsht.PrintStatus = defs.StatusFailed
```
boris-ilijic: Maybe should be better to use `PrintStatus` to have mapping logic in the single place.

cmd/pbm/status.go
```diff
@@ -638,6 +627,7 @@ func getStorageStat(
 			snpsht.Err = err
 			snpsht.ErrString = err.Error()
 			snpsht.Status = defs.StatusError
+			snpsht.PrintStatus = defs.StatusFailed
```
boris-ilijic: Maybe should be better to use PrintStatus.

boris-ilijic (dismissed):
Please see also comment.

boris-ilijic (approved):

PR #1146, opened 2025-06-12
PBM-1551 PBM reissues the previously executed command

veceraj:
Fixes previously executed command [PBM-1551](https://perconadev.atlassian.net/browse/PBM-1551): the agent now remembers every command it has seen, grouped by timestamp, and ignores repeats that share both ts and cmd. Old timestamps are purged as soon as we advance to a newer timestamp.

[PBM-1551]: https://perconadev.atlassian.net/browse/PBM-1551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pbm/ctrl/recv_test.go
+++ b/pbm/ctrl/recv_test.go
@@ -77,7 +77,7 @@ func TestListenCmd(t *testing.T) {
 			{"backup", 1},
 			{"restore", 1},
 			{"resync", 2},
-			{"backup", 3},
+			{"backup", 2},
 			{"restore", 2},
 		},
 	}


pbm/ctrl/recv_test.go
```diff
@@ -0,0 +1,142 @@
...
+		case cmd := <-cmdC:
+			got = append(got, cmd)
+		case err := <-errC:
+			t.Fatalf("listener error: %v", err)
+		}
+	}
+
+	if len(got) != 2 || got[0].Cmd != "restore" || got[1].Cmd != "backup" {
+		t.Errorf("duplicate leaked through: %v", got)
+	}
```
boris-ilijic, line 155: I would expand the same test with following cases:
- 3 commands within the same second
- 3 commands in T1, 2 commands in T2, 3 commands in T3

pbm/ctrl/recv_test.go
```diff
@@ -0,0 +1,234 @@
...
+		t.Fatalf("insert backup: %v", err)
+	}
+	if _, err := coll.InsertOne(ctx, testDoc); err != nil {
+		t.Fatalf("insert backup: %v", err)
+	}
+
+	cmdC, errC := ListenCmd(ctx, connClient, make(chan struct{}))
+
+	var got []Cmd
+	for len(got) < 2 {
```
boris-ilijic: we'll not detect potential duplicates if we expect just 2 msgs. I would have infinite loop, and jump out after 2-3 sec idle time.

pbm/ctrl/recv_test.go
```diff
@@ -0,0 +1,234 @@
...
+		case cmd := <-cmdC:
+			got = append(got, cmd)
+		case err := <-errC:
+			t.Fatalf("listener error: %v", err)
+		}
+	}
+
+	if len(got) != 2 || got[0].Cmd != "restore" || got[1].Cmd != "backup" {
+		t.Errorf("unexpected sequence: %v", got)
+	}
```
boris-ilijic: I would make one big test with all cases or use t.Run(...)

boris-ilijic (approved):

PR #1163, opened 2025-07-07
PBM-1488 PITR will stop saving the oplog if the process of copying the oplog from the backup fails

veceraj:
WIP


pbm/slicer/slicer.go
```diff
@@ -155,7 +155,6 @@ func (s *Slicer) Catchup(ctx context.Context) error {
 		err = s.copyReplsetOplog(ctx, rs)
 		if err != nil {
 			s.l.Error("copy oplog from %q backup: %v", lastBackup.Name, err)
-			return nil
```
boris-ilijic, line 158: This change will produce oplog hole within PITR chunks for the whole duration of the backup procedure.

boris-ilijic (changes requested):

PR #1183, opened 2025-08-29
PBM-1490: Backup with files that have reached remote storage limits

boris-ilijic:
[PBM-1490](https://jira.percona.com/browse/PBM-1490) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

This PR implements solution for handling backup files that exceed remote storage size limits by introducing a `split-merge` middleware. The middleware automatically splits large files into smaller parts during the backup procedure and merges them back during restore procedure. 

The file size threshold is set by default for each storage service, or it can be optionally set by configuration parameter:
| Storage        | Default Size | Config Param |
|-------------|--------------|--------------|
| AWS S3       |  4.9 TB           |  storage.s3.maxObjSizeGB |
| GCS             |  4.9 TB           |  storage.gcs.maxObjSizeGB |
| Azure Blob  |  190 TB             |  storage.azure.maxObjSizeGB            |
| File storage |  4.9 TB              |  storage.filesystem.maxObjSizeGB |

At the storage level, any file that exceeds the storage limit is named according to the following schema:
```
2025-09-04T09:09:47Z
│   └── rs1
│       ├── WiredTiger
│       ...
│       ├── collection-113-10993150527740990612.wt
│       ├── collection-119-10993150527740990612.wt
│       ├── collection-119-10993150527740990612.wt.pbmpart.1
│       ├── collection-119-10993150527740990612.wt.pbmpart.2
│       ├── collection-119-10993150527740990612.wt.pbmpart.3
│       ├── collection-119-10993150527740990612.wt.pbmpart.4
```

[PBM-1490]: https://perconadev.atlassian.net/browse/PBM-1490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pbm/storage/azure/azure.go
+++ b/pbm/storage/azure/azure.go
@@ -136,7 +136,7 @@ type Blob struct {
 	c *azblob.Client
 }
 
-func New(opts *Config, node string, l log.LogEvent) (*Blob, error) {
+func New(opts *Config, node string, l log.LogEvent) (storage.Storage, error) {
 	if l == nil {
 		l = log.DiscardEvent
 	}
@@ -152,7 +152,7 @@ func New(opts *Config, node string, l log.LogEvent) (*Blob, error) {
 		return nil, errors.Wrap(err, "init container")
 	}
 
-	return b, b.ensureContainer()
+	return storage.NewSplitMergeMW(b, opts.GetMaxObjSizeTB()), b.ensureContainer()
 }
 
 func (*Blob) Type() storage.Type {
--- a/pbm/storage/azure/azure.go
+++ b/pbm/storage/azure/azure.go
@@ -30,6 +30,8 @@ const (
 	defaultRetries = 10
 
 	maxBlocks = 50_000
+
+	DefaultMaxObjSizeTB = 190.0
 )
 
 //nolint:lll
@@ -40,6 +42,7 @@ type Config struct {
 	EndpointURLMap map[string]string `bson:"endpointUrlMap,omitempty" json:"endpointUrlMap,omitempty" yaml:"endpointUrlMap,omitempty"`
 	Prefix         string            `bson:"prefix" json:"prefix,omitempty" yaml:"prefix,omitempty"`
 	Credentials    Credentials       `bson:"credentials" json:"-" yaml:"credentials"`
+	MaxObjSizeTB   *float64          `bson:"maxObjSizeTB,omitempty" json:"maxObjSizeTB,omitempty" yaml:"maxObjSizeTB,omitempty"`
 }
 
 func (cfg *Config) Clone() *Config {
@@ -73,6 +76,9 @@ func (cfg *Config) Equal(other *Config) bool {
 		return false
 	}
 	if cfg.Credentials.Key != other.Credentials.Key {
+		return false
+	}
+	if cfg.MaxObjSizeTB != other.MaxObjSizeTB {
 		return false
 	}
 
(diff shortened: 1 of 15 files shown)


pbm/storage/split_merge_mw.go
```diff
@@ -0,0 +1,339 @@
...
+		err := sm.s.Save(fName, pr, options...)
+		if err != nil {
+			return errors.Wrap(err, "save during split-merge mw")
+		}
+
+		winfo := <-wInfoC
+		if winfo.err != nil {
+			if winfo.err == io.EOF && winfo.n != 0 {
+				break
+			} else if winfo.err == io.EOF && winfo.n == 0 {
```
inelpandzic, line 64: You could ditch this `else if` to remove additional indentation, and less indentations is always better. Like this:
```
		if winfo.err != nil {
			if winfo.err == io.EOF && winfo.n != 0 {
				break
			}

			if winfo.err == io.EOF && winfo.n == 0 {
				// empty part needs to be deleted, empty base should stay
				if isPartFile(fName) {
					if err := sm.s.Delete(fName); err != nil {
						return errors.Wrap(err, "empty file deletion")
					}
				}
				break
			}

			return errors.Wrap(winfo.err, "write pipeline split-merge mw")
		}

```

pbm/storage/azure/azure.go
```diff
@@ -30,6 +30,8 @@ const (
 	defaultRetries = 10
 
 	maxBlocks = 50_000
+
+	DefaultMaxObjSizeGB = 194560 // 190 TB
```
inelpandzic: Looks like this doesn't need to be exported. The same applies for other providers as well.

pbm/storage/split_merge_mw.go
```diff
@@ -0,0 +1,339 @@
...
+			return []FileInfo{}, errors.Wrap(err, "fetching next part")
+		}
+		res = append(res, fi)
+	}
+
+	return res, nil
+}
+
+// setPartsSize set pbm part size (in bytes) for the purpose of unit testing.
+func (sm *SplitMergeMiddleware) setPartsSize(maxObjSize int64) {
```
inelpandzic: Since `setPartsSize` and `getStorage` are only for unit tests, I would make them as functions and move them `split_merge_mw_test.go` that why you don't mix test code with the main one. So something like this:

```
func setSplitMergeMWPartsSize(sm *SplitMergeMiddleware, maxObjSize int64) {
	sm.maxObjSize = maxObjSize
}

func getSplitMergeMWStorage(sm *SplitMergeMiddleware) Storage {
	return sm.s
}
```

But also, this `getStorage` is used only once, I wouldn't have it at. And besides that I woudn't have this getter at all and would simply get the storage from the `s` field directly.

pbm/storage/split_merge_mw.go
```diff
@@ -0,0 +1,339 @@
...
+
+// GetBasePart extract base part of the file.
+// Base part is file without .pbmpart.xy suffix.
+func GetBasePart(fname string) string {
+	base := fname
+
+	pattern := regexp.MustCompile(`\.pbmpart\.\d+$`)
+	if pattern.MatchString(fname) {
+		fileParts := strings.Split(fname, ".")
+		base = strings.Join(fileParts[:len(fileParts)-2], ".")
```
inelpandzic: You can have this simpler:
```
		base = strings.Split(fname, ".pbmpart.")[0]
```
Or even better with `pbmPartToken` const:
```
		base = strings.Split(fname, pbmPartToken)[0]
```

pbm/storage/split_merge_mw.go
boris-ilijic: If we do not need `getStorage`, the same goes to `setPartsSize` or suggested `setSplitMergeMWPartsSize`. So within the tests it's ok to use fields directly. 
That was leftover from the one draft when both methods were public, afair. 

Will remove all of them and use fields within tests.

inelpandzic: Yes, you're right, even better that way.

PR #1186, opened 2025-09-05
PBM-1490: Improve storage.GetBasePart

inelpandzic:
[PBM-1490](https://jira.percona.com/browse/PBM-1490) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=percona&utm_campaign=badge_info)

PR improves `storage.GetBasePart` function by removing unnecessary regex compilation and matching which makes the implementation simpler and more performant.

[PBM-1490]: https://perconadev.atlassian.net/browse/PBM-1490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pbm/storage/split_merge_mw_test.go
+++ b/pbm/storage/split_merge_mw_test.go
@@ -165,7 +165,12 @@ func TestGetBasePart(t *testing.T) {
 		{
 			name:  "base part with pbmpart token, without index",
 			fname: "file_name.pbmpart.",
-			want:  "file_name",
+			want:  "file_name.pbmpart.",
+		},
+		{
+			name:  "base part with pbmpart token, invalid index",
+			fname: "file_name.pbmpart.23xx",
+			want:  "file_name.pbmpart.23xx",
 		},
 		{
 			name:  "pbmpart token exists, base part doesn't",
--- a/pbm/storage/split_merge_mw_test.go
+++ b/pbm/storage/split_merge_mw_test.go
@@ -144,3 +144,42 @@ func TestGetPartIndex(t *testing.T) {
 		}
 	})
 }
+
+// TestGetBasePart tests GetBasePart function.
+func TestGetBasePart(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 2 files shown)


pbm/storage/split_merge_mw_test.go
```diff
@@ -144,3 +144,42 @@ func TestGetPartIndex(t *testing.T) {
...
+		},
+		{
+			name:  "base part with pbmpart token and index",
+			fname: "file_name.pbmpart.15",
+			want:  "file_name",
+		},
+		{
+			name:  "base part with pbmpart token, without index",
+			fname: "file_name.pbmpart.",
+			want:  "file_name",
```
boris-ilijic: This is not valid pbm part base, want in this case should be:
```suggestion
			want:  "file_name.pbmpart."
```

inelpandzic: Aha, I see. Here is the fix [8780eac](https://github.com/percona/percona-backup-mongodb/pull/1186/commits/8780eac91b5bcaf894e11827fd2f673a562119ae)

It is not super simple as my original, but still simple enought and we avoid regex.

If you prefer that we have regex for some reason, then we should put the compilation part (`regexp.MustCompile('\.pbmpart\.\d+$')`) to a global variable, no need to compile it every time.

boris-ilijic: Yes, agree, let's keep regex and make it global, let's keep tests and PR is ready.

I prefer regex, because intention is clearer, and performance here does not make significant difference.

inelpandzic: I can agree with that. Here is the update: [f8f762c](https://github.com/percona/percona-backup-mongodb/pull/1186/commits/f8f762c7dc7923c1d2c95f0d4c1ad88167559d50)

PR #1203, opened 2025-10-01
PBM-1618: Minio as storage type for custom S3 type of storages

boris-ilijic:


pbm/storage/gcs/gcs.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 const (
+	GCSEndpointURL      = "storage.googleapis.com"
```
inelpandzic: Looks like it doesn't have to be exported.

inelpandzic (approved):

PR #1282, opened 2026-03-09
PBM-1598: Add configurable timeout for balancer stop command

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1598

The _configsvrBalancerStop admin command can block for hours when a chunk migration is in progress. This causes backups and restores to hang indefinitely while waiting for the balancer to stop.

Add a new backup.timeouts.balancerWait config parameter (in minutes) that limits how long PBM will wait for the balancer to stop. When the timeout is reached, the operation fails with a clear error. If unset or zero, PBM waits indefinitely (preserving current behavior).

### DIFF
--- a/pbm/config/config.go
+++ b/pbm/config/config.go
@@ -446,18 +446,18 @@ func (cfg *RestoreConf) Clone() *RestoreConf {
 //nolint:lll
 type RestoreTimeouts struct {
 	// BalancerStopMin is timeout (in minutes) to wait for the balancer to stop.
-	// 0 or nil means wait indefinitely (default).
-	BalancerStopMin *uint32 `bson:"balancerStopMin,omitempty" json:"balancerStopMin,omitempty" yaml:"balancerStopMin,omitempty"`
+	// 0 means wait indefinitely (default).
+	BalancerStopMin uint32 `bson:"balancerStopMin,omitempty" json:"balancerStopMin,omitempty" yaml:"balancerStopMin,omitempty"`
 }
 
 // BalancerStop returns timeout duration for waiting for the balancer to stop.
 // Returns 0 if not set, meaning PBM will wait indefinitely.
 func (t *RestoreTimeouts) BalancerStop() time.Duration {
-	if t == nil || t.BalancerStopMin == nil || *t.BalancerStopMin == 0 {
+	if t == nil {
 		return 0
 	}
 
-	return time.Duration(*t.BalancerStopMin) * time.Minute
+	return time.Duration(t.BalancerStopMin) * time.Minute
 }
 
 // GetFallbackEnabled gets config's or default value for fallbackEnabled
@@ -515,8 +515,8 @@ type BackupTimeouts struct {
 	Starting *uint32 `bson:"startingStatus,omitempty" json:"startingStatus,omitempty" yaml:"startingStatus,omitempty"`
 
 	// BalancerStopMin is timeout (in minutes) to wait for the balancer to stop.
-	// 0 or nil means wait indefinitely (default).
-	BalancerStopMin *uint32 `bson:"balancerStopMin,omitempty" json:"balancerStopMin,omitempty" yaml:"balancerStopMin,omitempty"`
+	// 0 means wait indefinitely (default).
+	BalancerStopMin uint32 `bson:"balancerStopMin,omitempty" json:"balancerStopMin,omitempty" yaml:"balancerStopMin,omitempty"`
 }
 
 // StartingStatus returns timeout duration for .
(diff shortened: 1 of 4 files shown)


pbm/config/config.go
```diff
@@ -489,6 +490,9 @@ func (cfg *BackupConf) Clone() *BackupConf {
 type BackupTimeouts struct {
 	// Starting is timeout (in seconds) to wait for a backup to start.
 	Starting *uint32 `bson:"startingStatus,omitempty" json:"startingStatus,omitempty" yaml:"startingStatus,omitempty"`
+	// BalancerWait is timeout (in minutes) to wait for the balancer to stop.
+	// If not set or zero, PBM will wait indefinitely
+	BalancerWait *uint32 `bson:"balancerWait,omitempty" json:"balancerWait,omitempty" yaml:"balancerWait,omitempty"`
```
jcechace: This might not be the race place for this config. Should we pull this up to top level? Considering it affects both -- backups and restores.

pbm/config/config.go
```diff
@@ -80,6 +80,10 @@ type Config struct {
 	Backup  *BackupConf  `bson:"backup,omitempty" json:"backup,omitempty" yaml:"backup,omitempty"`
 	Restore *RestoreConf `bson:"restore,omitempty" json:"restore,omitempty" yaml:"restore,omitempty"`
 
+	// BalancerWait is timeout (in minutes) to wait for the balancer to stop
+	// during backup or restore. 0 means wait indefinitely (default).
+	BalancerWait uint32 `bson:"balancerWait,omitempty" json:"balancerWait,omitempty" yaml:"balancerWait,omitempty"`
```
jcechace: Is this the right place for this config?

jcechace: Also currently this is in minutes (as per comment on the ticket), should we consider seconds? 

pro: It will be consistent with other waits (`e.g. backup.timeouts.startingStatus`)
con: User will almost always wan to set high number (e.g. `3600` for an hour instead of `60`)

boris-ilijic: > Is this the right place for this config?

We do not have any config param on the root level. I'd keep there just sections: backup, pitr, restore, storage, ...). In that case I am suggesting to have 2 parameters: one for backup and the other for restore. There are already something similar for:
https://docs.percona.com/percona-backup-mongodb/reference/backup-options.html#backupoplogspanmin
where we have it for backup and PITR.

> Also currently this is in minutes (as per comment on the ticket), should we consider seconds?

Probably both options are good. Personally I prefer minutes because that's right granularity for balancer imo. But seconds are also fine. Btw the upper parameter is in minutes (`backup.oplogSpanMin`), if we go with minutes we should add the suffix to be consistent.

boris-ilijic (commented):
Provided implementation adds timeout on the app level. Is it possible to pass timeout to existing internal command (`_configsvrBalancerStop`), something like this?
https://www.mongodb.com/docs/manual/reference/command/balancerStop/#syntax

pbm/config/config.go
jcechace: Ok, that works. Will adjust

jcechace:
> Provided implementation adds timeout on the app level. Is it possible to pass timeout to existing internal command (_configsvrBalancerStop), something like this?

Interesting, will check and test. I think we should also keep a timeout on the context, but I will look at this option as well.

jcechace:
Reworked to server-side timout passed over to `_configsvrBalancerStop` command (support confirmed). 

The configuration is now split and lives under `backup.timouts.balancerStopMin` and `restore.timeouts.balancerStopMin` respectively. 

As a consequence the timeout error is now cleaner 

```
Starting backup "2026-03-11T12:26:07Z".............Error: wait for backup status: set balancer OFF: run mongo command: (MaxTimeMSExpired) operation exceeded time limit
- Backup on replicaset "rs1" in state: error: waiting for running: cluster failed: set balancer OFF: run mongo command: (MaxTimeMSExpired) operation exceeded time limit
- Backup on replicaset "rs2" in state: error: waiting for running: cluster failed: set balancer OFF: run mongo command: (MaxTimeMSExpired) operation exceeded time limit
```

pbm/topo/cluster.go
```diff
@@ -219,8 +219,10 @@ func (b *BalancerStatus) IsDisabled() bool {
 	return b.Mode == BalancerModeOff && !b.InBalancerRound
 }
 
-// SetBalancerStatus sets balancer status
-func SetBalancerStatus(ctx context.Context, m connect.Client, mode BalancerMode) error {
+// SetBalancerStatus sets balancer status.
+// For BalancerModeOff, an optional maxTimeMS can be provided to limit how long
+// the server waits for the current balancer round to finish.
+func SetBalancerStatus(ctx context.Context, m connect.Client, mode BalancerMode, maxTimeMS ...int64) error {
```
boris-ilijic: Variadic param list is misleading here, I'd rather have pointer. Optionally, we can have second method if we don't want to pass nil (SetBalancerStatus + SetBalancerStatusWithTimeout).

boris-ilijic (dismissed):

jcechace:
`SetBalancerStatus` is the same as it was originally now, added `StopBalancer` -- the name is not symmetrical coz timeout is not supported for starting the balancer. 

Config property is now `int` instead of `*int` for simplicity. Otherwise nothing else changed.

boris-ilijic (approved):

PR #1283, opened 2026-03-10
PBM-1599: fix CLI timeout when balancer stop blocks

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1599

waitForBcpStatus now accepts StatusStarting so the CLI returns as soon as the backup is acknowledged by agents instead of waiting for StatusRunning (which requires balancer stop to complete first).

This was IMHO the cause of the error bellow (form the ticket). The code never got to executing `--wait` 

> Error: get backup metadata: get: context deadline exceeded
Starting backup '2025-08-12T00:30:03Z'..................................%

BTW: the ticket seems to present the output lines in wrong order. 

Also made the text output a bit nicer to make sense

### DIFF
--- a/cmd/pbm/backup.go
+++ b/cmd/pbm/backup.go
@@ -342,21 +342,7 @@ func waitForBcpStatus(ctx context.Context, conn connect.Client, bcpName string, 
 				return errors.Errorf("status error on %s", rs)
 			}
 		case <-ctx.Done():
-			if bmeta == nil {
-				return errors.New("no progress from leader, backup metadata not found")
-			}
-			rs := ""
-			for _, s := range bmeta.Replsets {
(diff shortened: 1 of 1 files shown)


boris-ilijic:
This will probably fix the bug from the ticket, but it'll also completely cut off all validations that we have before backup is started (with or w/o `-w` flag): 
https://docs.percona.com/percona-backup-mongodb/reference/backup-options.html?h=starting+status#backupcompressionlevel
I am not sure whether we are ready for that, but it wasn't planned in the ticket :) 

As a quick tip, maybe you can fix it by monitoring backups' heartbeats within the same select loop (didn't analyze the details, but it might be helpful).

PR #1284, opened 2026-03-12
PBM-1586, PBM-1648: Fix infinite cli wait on agent crush

jcechace:
Tickets:
- https://perconadev.atlassian.net/browse/PBM-1586
- https://perconadev.atlassian.net/browse/PBM-1648

pbm backup hangs forever when the backup leader agent crashes mid-backup because waitBackup() only checks the cluster-level status field, which nobody updates after the leader dies. This adds a stale heartbeat check — if the leader's heartbeat is older than 30 seconds, the CLI reports the backup as stuck with per-RS status detail, matching the detection already used by pbm status and agent-side waitForStatus.

```
dpbm backup -w
Starting backup "2026-03-12T10:16:39Z"....
Waiting for '2026-03-12T10:16:39Z' backup.............................. running
Error: backup stuck at "running" status, last heartbeat: 1773310600
- rs1: dumpDone
- rs2: dumpDone
- cfg: dumpDone
```

The PR also addresses a scenario when non-leader (shard) agent is the one that crushed -- which previously caused the leader to hang in waiting for cluster to converge to StatusDone

### DIFF
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -538,8 +538,8 @@ func (b *Backup) converged(
 				})
 
 				// nodes are cleaning its locks moving to the done status
-				// so no lock is ok and no need to ckech the heartbeats
-				if status != defs.StatusDone && !errors.Is(err, mongo.ErrNoDocuments) {
+				// so no lock is ok and no need to check the heartbeats
+				if shard.Status != defs.StatusDone && !errors.Is(err, mongo.ErrNoDocuments) {
 					if err != nil {
 						return false, errors.Wrapf(err, "unable to read lock for shard %s", shard.Name)
 					}
--- a/pbm/backup/backup.go
+++ b/pbm/backup/backup.go
@@ -537,9 +537,10 @@ func (b *Backup) converged(
 					Replset: shard.Name,
 				})
 
-				// nodes are cleaning its locks moving to the done status
-				// so no lock is ok and no need to check the heartbeats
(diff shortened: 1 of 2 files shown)


pbm/backup/backup.go
jcechace: @boris-ilijic  this covers the issue when a shard agent crushes and I think this change was the intended condition (based on the comment). 

For locks check we don't care about what state we are converging on, we care what state the shard is in. 


Now the condition says "If shard is not DONE and locks are not missing" -> check error and check lock's HB. 

**What this fixes:** a shard agent going down will not lead to backup stuck forever

**What this changes:** If HB goes stale on a backup in Canceled or Error state, the reported error will be "lost shard: ..." instead of `storage.ErrCancelled` or `"backup on shard %s failed with %s:..."`. 

The change can addressed and the fix improved by using 

```go
if shard.Status.IsRunning() && !errors.Is(err, mongo.ErrNoDocuments)
```

IMHO we care about locks only if the agent is not in terminal state ... So we could execute the entire lock query only for non terminal states.

In any case it's a minor difference impacting only what error would be returned (I think with `IsRunning()` it's a better one).

boris-ilijic: Thx for explanation, it makes sense, I'll analyze it more during final review.

boris-ilijic (commented):
PR looks good. Please finalize it and make it ready for review, and then I'll do final review.

boris-ilijic (approved):

PR #1286, opened 2026-03-16
PBM-1599: CLI fails with timeout when stopping the balancer takes longer

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1599

Changes made:

- Running `pbm backup` now: 
    - first waits up to `defs.WaitBackupStart` to confirm backup meta vas saved
    - then waits for backup to reach StatusRunning without any timeout
- Running `pbm backup -w` remains unchanged 
- minor cosmetic tweak to print errors on new lines consistently with and w/o `-w`

### DIFF
--- a/cmd/pbm/backup.go
+++ b/cmd/pbm/backup.go
@@ -342,21 +342,7 @@ func waitForBcpStatus(ctx context.Context, conn connect.Client, bcpName string, 
 				return errors.Errorf("status error on %s", rs)
 			}
 		case <-ctx.Done():
-			if bmeta == nil {
-				return errors.New("no progress from leader, backup metadata not found")
-			}
-			rs := ""
-			for _, s := range bmeta.Replsets {
-				rs += fmt.Sprintf("- Backup on replicaset \"%s\" in state: %v\n", s.Name, s.Status)
-				if s.Error != "" {
-					rs += ": " + s.Error
-				}
-			}
-			if rs == "" {
-				rs = "<no replset has started backup>\n"
-			}
-
-			return errors.New("no confirmation that backup has successfully started. Replsets status:\n" + rs)
+			return ctx.Err()
 		}
 	}
 }
--- a/cmd/pbm/backup.go
+++ b/cmd/pbm/backup.go
@@ -316,15 +316,13 @@ func waitForBcpStatus(ctx context.Context, conn connect.Client, bcpName string, 
 	tk := time.NewTicker(time.Second)
 	defer tk.Stop()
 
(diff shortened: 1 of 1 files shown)


jcechace:
@boris-ilijic If you agree I will handle the edge case for stuck backup as part of [PBM-1586](https://perconadev.atlassian.net/browse/PBM-1586)

[PBM-1586]: https://perconadev.atlassian.net/browse/PBM-1586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

cmd/pbm/backup.go
```diff
@@ -316,21 +345,7 @@ func waitForBcpStatus(ctx context.Context, conn connect.Client, bcpName string,
...
-				}
-			}
-			if rs == "" {
-				rs = "<no replset has started backup>\n"
-			}
-
-			return errors.New("no confirmation that backup has successfully started. Replsets status:\n" + rs)
+			return ctx.Err()
 		}
 	}
```
jcechace, line 335: @boris-ilijic I think w/o timeout we don't need this branch and can just return ctx.Err() consistently with waitBackup. This can happen only on external cancelation (user killing the CLI in this case).

cmd/pbm/backup.go
```diff
@@ -280,25 +281,51 @@ func waitBackup(
...
 }
 
-func waitForBcpStatus(ctx context.Context, conn connect.Client, bcpName string, showProgress bool) error {
+func waitForBcpExists(ctx context.Context, conn connect.Client, bcpName string, showProgress bool) error {
 	tk := time.NewTicker(time.Second)
 	defer tk.Stop()
 
-	var bmeta *backup.BackupMeta
+	existCtx, cancel := context.WithTimeout(ctx, defs.WaitBackupStart)
+	defer cancel()
```
boris-ilijic:
```
to := time.After(defs.WaitBackupStart)
```
is slightly common pattern in this case. And within select statement:
```
case <-to:
```

cmd/pbm/backup.go
boris-ilijic: Yes, let's handle that as part of the other ticket.

boris-ilijic (dismissed):

cmd/pbm/backup.go
jcechace: Cool, actually didn't know that one! Thanks.

boris-ilijic (approved):

PR #1287, opened 2026-03-17
PBM-1167: PiTR with physical restore doesn't recover data due to rollback

boris-ilijic:
PR for https://perconadev.atlassian.net/browse/PBM-1167

Currently, when executing physical restore with PITR, PBM applies PITR entries only on the Primary node within RS, while Secondary nodes contain just restored data without PITR. That approach implies that former Primary node needs to be started first after the restore procedure is done. Secondary nodes will do catchup, after they are started.

This PR improves physical restore with PITR  through several changes:
- After the restore with PITR, it's possible to start RS nodes in any order.
- PITR entries are applied to all nodes, not just ex-Primary. Using that approach, all nodes have the same data files after the restore is done.
- To apply PITR to every node and ensure an equal oplog history after the restore, PBM applies PITR entries to a standalone instance.


pbm/restore/physical.go
```diff
@@ -1754,62 +1754,24 @@ func (r *PhysRestore) recoverStandalone() error {
 	return r.shutdown(c)
 }
 
-func (r *PhysRestore) replayOplog(
+func (r *PhysRestore) replayPITROnStandalone(
```
jcechace, line 1757: A nitpick. I think `replayOplog` or `replayOplogOnStandalone`  or `applyOplogOnStandalone` would be a better name...  After all we are applying/replaying oplog entries. 

"Replay point in time recovery..." sounds kinda strange. 

Nothing major though, and I'm fine keeping it as is.

jcechace (approved):
Minor naming comment, up to you what we go with. 

Thanks, I learned a bunch on this one ;).

pbm/restore/physical.go
boris-ilijic: That's valid concern, and within git history there's exact the name that you proposed (`replayOplogOnStandalone`) and then I renamed it with this [commit](https://github.com/percona/percona-backup-mongodb/pull/1287/changes/9131994022f651906c21119e001244830ee77361#diff-d99d228622352026338b852f0e347ffa7e403639e5460fbb6205ec955a98b197L1384).
The potential issue with that name is that in this use case we are dealing with oplog from the db (rs.oplog) as well as oplog from PITR, and I wanted to distinguish those two. 

Additionally, everything related to term PITR sounds strange (e.g. backup PITR(ecovery), or restore PITR), we had lots of those discussions in the past, but we decided to treat that just as the feature name.

PR #1302, opened 2026-04-08
PBM-1730 Abort restore on balancer stop and handle timeouts

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1730

- Reintroduces `restore.timeouts.stopBalancer`
- Balancer is stopped only for logical restore 
- Balancer is stopped after restore meta exists (right after `r.init()` call)
- Failure to stop the balancer leads to a backup failure and stop
- `pbm restore` waiting behavior unified with `pbm backup`
   -  without `--wait` wait for metada existence with timout, then unconditionally wait for `StatusRunning` while monitoring HB 
   - with `--wait` wait for terminal status while monitoring HB (behavior unchanged) 
   
Wait difference for `json` ouput:
- CLI doesn't wait for `StatusRunning` when ouput is not text
- `pbm restore <name> --out=json` is basically "fire and forget", unlike backup, which waits for `StatusRunning` regardless of output format
- This is a current dev behavior which I kept

Last two commits are pure refactoring, which IMHO makes the code more understandable but could be dropped entirely

### DIFF
--- a/cmd/pbm/oplog.go
+++ b/cmd/pbm/oplog.go
@@ -104,7 +104,7 @@ func replayOplog(
 	startCtx, cancel := context.WithTimeout(ctx, defs.WaitActionStart)
 	defer cancel()
 
-	m, err := waitForRestoreStatus(startCtx, conn, name, restore.GetRestoreMeta)
+	m, err := waitForReplayStatus(startCtx, conn, name, restore.GetRestoreMeta)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pbm/restore/logical.go
```diff
@@ -144,6 +144,40 @@ func (r *Restore) exit(ctx context.Context, err error) {
...
+			return errors.Wrap(err, "set balancer off")
+		}
+
+		r.log.Debug("waiting for balancer off")
+		bs := topo.WaitForBalancerDisabled(ctx, r.leadConn, time.Second*30, r.log)
+		if bs.IsDisabled() {
+			r.log.Debug("balancer is disabled")
+		} else {
+			r.log.Warning("balancer is not disabled: balancer mode: %s, in balancer round: %t",
+				bs.Mode, bs.InBalancerRound)
```
jcechace, line 174: Should this remain a warning? It is also a warning for backup. 
Happne when balancer was disabled but is still in an active round.

cmd/pbm/oplog.go
```diff
@@ -104,7 +104,7 @@ func replayOplog(
 	startCtx, cancel := context.WithTimeout(ctx, defs.WaitActionStart)
 	defer cancel()
 
-	m, err := waitForRestoreStatus(startCtx, conn, name, restore.GetRestoreMeta)
+	m, err := waitForReplayStatus(startCtx, conn, name, restore.GetRestoreMeta)
```
jcechace: Depends whether we want to change the wait logic to unbounded also for PITR replay. I decided to keep the change isolated to restores.

jcechace: Unified, the same `waitForRestoreStatus` is used

boris-ilijic (approved):

PR #1314, opened 2026-05-01
Bump mongo-go-driver: v1.17.9 -> v2.6.0

boris-ilijic:
PR for https://perconadev.atlassian.net/browse/PBM-1592

PR migrates from go.mongodb.org/mongo-driver v1 to v2 (v2.6.0) and bumps the related `mongo-tools` dependency to latest.
Follow up tasks related `mongo-tools` will be part of: https://perconadev.atlassian.net/browse/PBM-1682.

```
┌───────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────┐
│                          Old dependency                           │                          New dependency                           │
├───────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ go.mongodb.org/mongo-driver v1.17.9                               │ go.mongodb.org/mongo-driver/v2 v2.6.0                             │
├───────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ github.com/mongodb/mongo-tools v0.0.0-20240723193119-837c2bc263f4 │ github.com/mongodb/mongo-tools v0.0.0-20260424130528-bbe4099afe32 │
├───────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ go 1.25.5                                                         │ go 1.25.9                                                         │
└───────────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────┘
```


pbm/connect/connect.go
```diff
@@ -364,8 +368,11 @@ func (l *clientImpl) applyOptonsFromConnString(cmd bson.D) bson.D {
 	cmdName := cmd[0].Key
 	switch cmdName {
 	case "create":
-		if l.options.WriteConcern != nil {
-			cmd = append(cmd, bson.E{"writeConcern", l.options.WriteConcern})
+		if wc := l.options.WriteConcern; wc != nil && wc.W != nil {
+			cmd = append(cmd, bson.E{
+				Key:   "writeConcern",
+				Value: bson.D{{Key: "w", Value: wc.W}},
+			})
```
jcechace, line 375: Is this intentionally ignoring ignoring `wc.Journal`?

pbm/archive/archive.go
```diff
@@ -224,7 +224,8 @@ func closeChunk(w io.Writer, ns *Namespace) error {
 		Database:   ns.Database,
 		Collection: ns.Collection,
 		EOF:        true,
-		CRC:        ns.CRC,
+		// migration mongo-driver v1->v2: fix this in PBM-1682
+		CRC: uint64(ns.CRC), //nolint:gosec
```
jcechace, line 228: Just for information. What needs to be fixed here? On binary level this should be safe and it looks like we are using `crc64.ECMA` for CRC calculations.

pbm/oplog/restore.go
```diff
@@ -915,7 +915,8 @@ func txnInnerOps(txnOp *db.Oplog) ([]db.Oplog, error) {
 		// so we are assigning them from the parent transaction op
 		op.Timestamp = txnOp.Timestamp
 		op.Term = txnOp.Term
-		op.Hash = txnOp.Hash
+		// migration mongo-driver v1->v2: fix this in PBM-1682
+		// op.Hash = txnOp.Hash
```
jcechace, line 919: Just for info. What are the implications of dropping this? Seems that field was last useful in MongoDB 4.0

jcechace (approved):
LGTM. Left few questions, mostly just out of curiosity.

pbm/connect/connect.go
boris-ilijic: This is not connection object, this is `create` command that also has `writeConcern` property. 
See here: https://www.mongodb.com/docs/manual/reference/command/create/#syntax
So `j` is not relevant here.

pbm/archive/archive.go
boris-ilijic: I'll double check as part of PBM-1682

pbm/oplog/restore.go
boris-ilijic: Yes, agree, this is obsolete but I'll double check impact.

PR #1315, opened 2026-05-04
PBM-1668 Index commit quorum for logical restore

jcechace:
ticket: https://perconadev.atlassian.net/browse/PBM-1668

This change lets users control the commit quorum used when PBM recreates indexes during logical restore.
Users can configure the behavior globally with restore.indexCommitQuorum, or override it for a single restore with pbm restore --index-commit-quorum. Supported values are:
- majority
- votingMembers
- a non-negative integer node count
If no value is provided, PBM uses majority.

### DIFF
--- a/pbm/defs/defs.go
+++ b/pbm/defs/defs.go
@@ -211,7 +211,7 @@ const (
 	IndexCommitQuorumVotingMembers IndexCommitQuorum = "votingMembers"
 )
 
-const DefaultRestoreIndexCommitQuorum = IndexCommitQuorumMajority
+const DefaultRestoreIndexCommitQuorum = IndexCommitQuorumVotingMembers
 
 // CommandValue returns the value shape expected by MongoDB's createIndexes
 // commitQuorum field. MongoDB accepts symbolic string values or an integer
--- a/pbm/defs/defs.go
+++ b/pbm/defs/defs.go
@@ -1,8 +1,6 @@
 package defs
 
 import (
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/percona/percona-backup-mongodb/pbm/compress"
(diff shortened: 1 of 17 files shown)


jcechace:
@boris-ilijic  `0` is currently not allowed as a value, although it is valid. From MongoDB docs

>0 - Disables quorum-voting behavior. Members start the index build simultaneously but do not vote or wait for quorum before completing the index build. If you start an index build with a commit quorum of 0, you cannot later modify the commit quorum using [setIndexCommitQuorum.](https://www.mongodb.com/docs/manual/reference/command/setIndexCommitQuorum/#mongodb-dbcommand-dbcmd.setIndexCommitQuorum)

Do we want to support it?

boris-ilijic:
> @boris-ilijic `0` is currently not allowed as a value, although it is valid. From MongoDB docs
> 
> > 0 - Disables quorum-voting behavior. Members start the index build simultaneously but do not vote or wait for quorum before completing the index build. If you start an index build with a commit quorum of 0, you cannot later modify the commit quorum using [setIndexCommitQuorum.](https://www.mongodb.com/docs/manual/reference/command/setIndexCommitQuorum/#mongodb-dbcommand-dbcmd.setIndexCommitQuorum)
> 
> Do we want to support it?

Imo, we should, that mode will unblock the restore, and then DBA can monitor separately what's happening on the secondaries and what's lag is there. But maybe we should test that a bit more.

pbm/defs/defs.go
```diff
@@ -201,3 +203,43 @@ const (
...
+		return nil
+	}
+
+	n, err := strconv.ParseInt(val, 10, 32)
+	if err != nil || n <= 0 {
+		return errors.Errorf("invalid restore.indexCommitQuorum %q", val)
+	}
+
+	return nil
+}
```
boris-ilijic: It's preferable to remove this logic from `defs` package. E.g. how it was done for `compress` package is approach to follow.

boris-ilijic (approved):

PR #1321, opened 2026-05-20
PBM-1728 Support workload federation for OCI

jcechace:
ticket: https://perconadev.atlassian.net/browse/PBM-1728
related to: https://github.com/percona/percona-backup-mongodb/pull/1318

- Adds support for instance principal auth
- Adds support for OKE workload identity
- Credentials shape changed to accommodate the different types of authentication

Example configuration 

```yaml
storage:
  type: oci
  oci:
    region: eu-frankfurt-1
    namespace: idvufsl0apl6
    bucket: pbm-test-jakub
    prefix: pbm
    retryer:
      maxAttempts: 15
      maxBackoff: 180s
    credentials:
      type: userPrincipal # instancePrincipal | okeWorkloadIdentity
      userPrincipal:
        tenancy: ocid1.tenancy.oc1..xxx
        user: ocid1.user.oc1..yyy
        fingerprint: e0:a8:24:e5:f6:b8:6e:...
        privateKey: |
          -----BEGIN PRIVATE KEY-----

          ...
          -----END PRIVATE KEY-----
```

PR #1318, opened 2026-05-14
PBM-1727 Support OCI storage

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1727

- Adds native OCI Object Storage backend using the OCI Go SDK, not S3 compatibility.
- Supports inline OCI user-principal credentials, namespace/bucket/prefix config, and PBM storage wiring.
- Implements OCI storage operations: upload, ranged download, stat, list, delete, and native server-side copy.
- Adds native CopyObject support with work-request polling, error detail reporting, and bounded polling timeout.
- Adds OCI retry configuration for SDK operations and transfer-manager uploads.
- Adds concurrent/ranged downloader support using PBM’s shared storage.PartReader.
- Includes focused unit tests for config validation, retry behavior, native copy, copy timeout, and ranged downloader behavior.

PR #1322, opened 2026-05-25
PBM-1727 OCI SSE customization support + empty upload fix 

jcechace:
**Ticket:**
https://perconadev.atlassian.net/browse/PBM-1727

**Follows on:**
https://github.com/percona/percona-backup-mongodb/pull/1321
https://github.com/percona/percona-backup-mongodb/pull/1318

**Description:**
This PR  includes the following changes

- Support for SSE related configuration. 
  - SSE with KMS by setting key ID 
  - SSE-C by setting a customer AES256 key 
  -  Note: if SSE is not configured, OCI always encrypts stored files via default MEK. 
- OCI Integration tests against live OCI (gated behind build tag)
- Fixes for resulting from these addition
  - FileStat now returns the correct result for empty files
  - Workaround for saving and encrypting empty objects (OCI SDK bugs)
    -  https://github.com/oracle/oci-go-sdk/issues/648
    -  https://github.com/oracle/oci-go-sdk/issues/647

**Example config:**

```yaml
storage:
  type: oci
  oci:
    region: eu-frankfurt-1
    namespace: ...
    bucket: pbm-test-jakub
    prefix: pbm
    retryer:
      maxAttempts: 15
      maxBackoff: 180s
    serverSideEncryption:
      # only one of following allowed
      sseCustomerKey: 5dQiQ...dKviuc6q2Q4k=
      kmsKeyID: ocid1.key.oc1.eu-frankfurt-1.envbichxaaexg....
    credentials:
      type: userPrincipal
      userPrincipal:
        tenancy: ocid1.tenancy.oc1.....
        user: ocid1.user.oc1.....
        fingerprint: e0:a8:24:e5:f6:b8:...
        privateKey: |
          -----BEGIN PRIVATE KEY-----
          ...
          -----END PRIVATE KEY-----
```


pbm/storage/oci/config.go
```diff
@@ -43,6 +48,8 @@ type Config struct {
 	Credentials Credentials `bson:"credentials" json:"credentials" yaml:"credentials"`
 	Retryer     *Retryer    `bson:"retryer,omitempty" json:"retryer,omitempty" yaml:"retryer,omitempty"`
 
+	ServerSideEncryption SSE `bson:"serverSideEncryption,omitempty" json:"serverSideEncryption,omitempty" yaml:"serverSideEncryption,omitempty"`
```
boris-ilijic: omitempty doesn't have effect here, let's remove it

pbm/storage/oci/config.go
```diff
@@ -59,6 +66,45 @@ type Retryer struct {
...
+		return sseHeaders{}, errors.Wrap(err, "decode sseCustomerKey")
+	}
+	sum := sha256.Sum256(decodedKey)
+
+	return sseHeaders{
+		customerAlgorithm: common.String(sseCustomerAlgorithm),
+		customerKey:       common.String(string(s.SseCustomerKey)),
+		customerKeySHA256: common.String(base64.StdEncoding.EncodeToString(sum[:])),
+	}, nil
+}
```
boris-ilijic: I'd put it into `oci.go` or separate file, because in future we might extract this config into package `config`. That also means to unbind headers() method from SSE type.

pbm/storage/oci/config.go
```diff
@@ -1,9 +1,13 @@
 package oci
 
 import (
+	"crypto/sha256"
+	"encoding/base64"
 	"reflect"
 	"time"
 
+	"github.com/oracle/oci-go-sdk/v65/common"
```
boris-ilijic: It should be preferable to not have provider dependency within config.go

boris-ilijic (dismissed):

pbm/storage/oci/config.go
jcechace: Looked into it. For YAML it has effect, for BSON we would have to implement IsZero(), nothing to be done for JSON. 

So I will remove it, marshaling into empty value is acceptable -- coz it just means "no customization" since SSE is always enabled

boris-ilijic (approved):

PR #1325, opened 2026-05-28
PBM-1710 PBM add SBOM into packaging

vorsel:
- Install Syft and Grype.
- Generate a CycloneDX 1.6 SBOM for the binary tarball
  Scope is limited to the 4 built PBM Go binaries in ./bin folder.
  using the go-module-binary cataloger; file catalogers are disabled to
  keep package-level granularity. SBOM is placed at the tarball as
  <product>-<version>.cdx.json file.
- Inline syft in debian/rules and rpm spec to produce a CycloneDX 1.6 SBOM
for the package's Go binaries. SBOM is placed at
/usr/share/doc/percona-backup-mongodb/<product>-<version>.cdx.json.
- Patch syft output via jq so metadata.component is a proper application
identity (type=application, human-readable bom-ref, purl) instead of
syft's default type=file with an opaque hash bom-ref. PURL type matches
the artifact: pkg:generic for tarball, pkg:deb for DEB, pkg:rpm for RPM.

### DIFF
--- a/packaging/debian/install
+++ b/packaging/debian/install
@@ -13,3 +13,4 @@ completions/zsh/_pbm-agent /usr/share/zsh/vendor-completions/
 completions/zsh/_pbm /usr/share/zsh/vendor-completions/
 completions/zsh/_pbm-speed-test /usr/share/zsh/vendor-completions/
 LICENSE /usr/share/doc/percona-backup-mongodb/
+percona-backup-mongodb-*.cdx.json /usr/share/doc/percona-backup-mongodb/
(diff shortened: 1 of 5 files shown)


nogueiraanderson:
LGTM. One nit, the `local TARBALL=$1` / `local param=$1` hardening you applied in PCSM 238 (`get_tar` and `get_deb_sources`) is missing from `mongodb-backup_builder.sh`. PBM's main flow runs `build_tarball` before the deb/rpm stages so the shadowing bug doesn't fire today, but matching the sibling now avoids paper cuts the next time someone reorders that script. Optional but cheap.

nogueiraanderson (approved):
LGTM. See inline comment for the optional nit.

boris-ilijic (approved):

PR #1326, opened 2026-05-29
PBM-1715: Multi-files parallel upload and download during physical backup/restore with FS/NFS

boris-ilijic:
PR for https://perconadev.atlassian.net/browse/PBM-1715

PR adds support for parallel file upload/download during backup/physical restore for FS/NFS storage types.

The following configuration parameters and options are available:
```
$ pbm config --list
storage:
  type: filesystem
  filesystem:
    path: /opt/backups/pbm
backup:
  numParallelFiles: 4
restore:
  numParallelFiles: 4
```

```
$ pbm backup --help
Make backup

Usage:
  pbm backup [flags]

Flags:
....
      --num-parallel-files int32        Number of files to upload in parallel (physical backup, filesystem storage only)
```
```
$ pbm restore --help
Restore backup

Usage:
  pbm restore [backup_name] [flags]

Flags:
....
      --num-parallel-files int32       Number of files to copy in parallel (physical restore, filesystem storage only)
```

### DIFF
--- a/pbm/backup/physical.go
+++ b/pbm/backup/physical.go
@@ -658,12 +658,61 @@ func trimFilePrefix(fname, trimPrefix string) string {
 	return path.Clean("./" + strings.TrimPrefix(fname, trimPrefix))
 }
 
-// Uploads given files to the storage. files may come as 16Mb (by default)
-// blocks in that case it will concat consecutive blocks in one bigger file.
+// upItem is one entry of an upload.
+// It contains the file name and and info about whether it's necessary to upload it.
+type upItem struct {
+	file   File
+	upload bool
+}
+
+// planUploads walks files in order and produces the upload plan. files may
+// come as 16Mb (by default) blocks; in that case consecutive blocks of the
+// same file are coalesced into one bigger upload.
 // For example: f1[0-16], f1[16-24], f1[64-16] becomes f1[0-24], f1[50-16].
-// If this is an incremental, NOT base backup, it will skip uploading of
-// unchanged files (Len == 0) but add them to the meta as we need know
-// what files shouldn't be restored (those which isn't in the target backup).
+// If this is an incremental, NOT base backup, unchanged files (Len == 0) are
+// not uploaded but still recorded in the meta as we need to know what files
+// shouldn't be restored (those which aren't in the target backup).
+func planUploads(files []File, incr bool) []upItem {
+	if len(files) == 0 {
+		return nil
+	}
+
+	upItems := make([]upItem, 0, len(files))
+	wfile := files[0]
+	for _, file := range files[1:] {
+		// Skip uploading unchanged files if incremental
+		// but add them to the meta to keep track of files to be restored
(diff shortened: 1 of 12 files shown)


pbm/restore/physical.go
```diff
@@ -1631,21 +1652,87 @@ func (r *PhysRestore) copyFiles() (*storage.DownloadStat, error) {
...
+	}
+	// each concurrent copy needs its own buffer
+	bufPool := make(chan []byte, numWorkers)
+	allBufs := make([]byte, numWorkers*bufLen)
+	for i := range numWorkers {
+		// preallocate buffer for each worker
+		bufPool <- allBufs[i*bufLen : (i+1)*bufLen : (i+1)*bufLen]
+	}
+
+	eg, egCtx := errgroup.WithContext(context.Background())
```
jcechace, line 1703: would it makes sense for `r.copyFiles` to receive `ctx` from `Snapshot` and base context on that?  Then later do

```go
for _, op := range job.ops {
    if err := egCtx.Err(); err != nil {
        return err
    }

    if err := r.copyFile(op.src, job.dst, op.fMeta, op.cmpr, cpBuf); err != nil {
        return err
    }
}
```

that would allow for a faster cancelation of running jobs (basically multi op jobs could cancel mid run). 

But it may not be neccessary

pbm/backup/physical.go
```diff
@@ -710,32 +709,99 @@ func uploadFiles(
...
 	}
 
-	f, err := writeFile(ctx, &wfile, path.Join(subdir, trim(wfile.Name)), stg, comprT, comprL,
-		cpBuf, saveBuf, fsSaveBuf)
-	if err != nil {
-		return data, errors.Wrapf(err, "upload file `%s`", wfile.Name)
+	// each goroutine writes a distinct index, so no locking needed.
+	results := make([]File, len(upItems))
+	eg, egCtx := errgroup.WithContext(ctx)
+	eg.SetLimit(numWorkers)
```
jcechace: are we OK doing this unbounded?  `numWorkers` can currently eb whatever user configures it to be, we then allocate buffers and everything based on that. Maybe we should set some upper bound and then use `numWorkers = min(numWorkers, len(upItems))`

equally for restores

jcechace (dismissed):

pbm/backup/physical.go
boris-ilijic: I was thinking to introduce fixed cap, but at the same time we want to allow users to experiment. Currently it starts one goroutine for each item (`range upItems`) so there's already bounded behavior, but it might be beneficial in case of some future refactoring, so I'll add it explicitly. Also, it clarifies the intention.

pbm/restore/physical.go
boris-ilijic: Improved in on the job level, and made it to fail faster. It's not necessary to propagate ctx within copyFile because majority of work is done within CopyBuffer, and do not want to optimize it more because we are using the same logic everywhere.

jcechace (approved):

PR #1334, opened 2026-06-18
PBM-1675: Disallow connections with mongos URI

jcechace:
**Ticket:** https://perconadev.atlassian.net/browse/PBM-1675
**Follows up on:** https://github.com/percona/percona-backup-mongodb/pull/1311

~~Technically the issue fixed by this PR is unrelated to the original ticket. Only after QE tested with mongos connection on 7.0 was this discovered.~~

**Description:**
~~The new `topo.GetClusterFCV` calls `version.GetFCV`. In case of success it's done. On failure, it checks whether we are connected through mongos, and if so, creates a temporary connection to CSRS directly and resolves FCV against that.~~

After a discussion and more investigation, we found more potential problem surface for mongos connection. As a result a decision was made to disallow mongos connection altogether. This PR thus modifies `connect.Connect()` to return an error when mongos URI is used (for both, cli and agent -- which already had a secondary rejection path).

```sh
pbm --mongodb-uri mongodb://dba:test1234@mongos:27017 status
Error: connect to mongodb: mongos connection is not supported
```


jcechace:
@boris-ilijic  Do you think this approach is OK? I would lean towards yes, but just very slightly. I can see the appeal in using mongos connection from CLI, so I would like to keep that option. But it's true that this adds complexity.

boris-ilijic:
The intention and code are clear, and that's good. But I'd apply here KISS principle and go with another mentioned expected behavior from the ticket:
> Expected behavior: either show warning that URI is incorrect

Main reasons:
- monogs is down during restore, which means that status will not work in such cases, and that will not be nice UX. From docs:
> Shut down all mongos nodes to stop clients from accessing the database while restore is in progress.

- another connection is created for every `pbm status` call, and some customers are polling that cli cmd using scripts, so that will influence the number of connections in such cases.
- maintainability, as you mentioned.

pbm/connect/connect.go
```diff
@@ -273,24 +272,6 @@ func Connect(ctx context.Context, uri, appName string) (*clientImpl, error) {
...
-		return errors.Wrap(err, "get node info ext")
-	}
-
-	if !info.isMongos() && !info.isClusterLeader() {
-		return ErrInvalidConnection
-	}
-
-	return nil
-}
-
```
jcechace, line 293: This method was a dead code

boris-ilijic (approved):

PR #703, opened 2022-09-06
separate storage for each file

defbin:

PR #1344, opened 2026-06-30
PBM-1776 GCS parallel upload (with storage close)

jcechace:
```yaml
storage:
  type: gcs
  gcs:
    bucket: pbm-jakub
    prefix: pbm
    maxObjSizeGB: 1
    credentials:
      clientEmail: "pbm-jakub@dev-sandbox-348314.iam.gserviceaccount.com"
      privateKey: "..."
      clientType: grpc # json | grpc 
      parallelUploadConcurrency: 4 # number of concurrent parts, enabled requires > 1
 ```
 
 Status will remain experimental due to ongoing work in GCS. 
 At least these need to be released:
 - https://github.com/googleapis/google-cloud-go/pull/14704
 - https://github.com/googleapis/google-cloud-go/pull/19930

### DIFF
--- a/pbm/restore/logical.go
+++ b/pbm/restore/logical.go
@@ -134,6 +134,12 @@ func (r *Restore) Close() {
 	if r.stopHB != nil {
 		close(r.stopHB)
 	}
+
+	storage.Close(r.bcpStg, r.log)
+	r.bcpStg = nil
+
+	storage.Close(r.oplogStg, r.log)
+	r.oplogStg = nil
 }
 
 func (r *Restore) exit(ctx context.Context, err error) {
--- a/pbm/restore/logical.go
+++ b/pbm/restore/logical.go
@@ -752,11 +752,13 @@ func LookupBackupMeta(
 		return nil, errors.Wrap(err, "get backup metadata from db")
 	}
 
+	l := log.LogEventFromContext(ctx)
 	var stg storage.Storage
-	stg, err = util.GetStorage(ctx, conn, node, log.LogEventFromContext(ctx))
+	stg, err = util.GetStorage(ctx, conn, node, l)
 	if err != nil {
 		return nil, errors.Wrap(err, "get storage")
 	}
+	defer storage.Close(stg, l)
 
 	bcp, err = GetMetaFromStore(stg, backupName)
 	if err != nil {
--- a/pbm/restore/logical.go
+++ b/pbm/restore/logical.go
@@ -87,6 +87,22 @@ type oplogRange struct {
 	storage storage.Storage
 }
 
+type storageReadCloser struct {
+	r   io.ReadCloser
+	stg storage.Storage
+	l   log.LogEvent
+}
+
+func (rc storageReadCloser) Read(p []byte) (int, error) {
+	return rc.r.Read(p)
+}
+
+func (rc storageReadCloser) Close() error {
+	err := rc.r.Close()
+	storage.Close(rc.stg, rc.l)
+	return err
+}
+
 // configDatabasesDoc represents document in config.databases collection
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: pbm/storage/gcs/config.go [hunk 1]
```diff
@@ -8,20 +8,48 @@ import (
...
 	Retryer *Retryer `bson:"retryer,omitempty" json:"retryer,omitempty" yaml:"retryer,omitempty"`
 }
 
+// ParallelUpload configures the experimental Google Cloud Storage parallel upload feature.
+//
+// Parallel uploads are supported only by the Google SDK gRPC client.
+// HMAC credentials continue to use the MinIO client and are not affected by this setting.
+type ParallelUpload struct {
+	// Enabled enables Writer parallel uploads when clientType is grpc.
+	Enabled bool `bson:"enabled,omitempty" json:"enabled,omitempty" yaml:"enabled,omitempty"`
```
boris-ilijic: This is redundant parameter, we can disable/enable it with MaxConcurrency

pbm/storage/gcs/config.go
```diff
@@ -8,20 +8,48 @@ import (
...
+
+	// PartSize is the size of each temporary part in bytes.
+	// If unset, PBM uses the computed ChunkSize for the uploaded object.
+	PartSize int `bson:"partSize,omitempty" json:"partSize,omitempty" yaml:"partSize,omitempty"`
```
boris-ilijic: ChunkSize can be reused here, or maybe even better use PartSize for both (json and grpc) but in that case it should be out of ParallelUpload block.

pbm/storage/gcs/google_client.go
```diff
@@ -137,31 +156,57 @@ func (g googleClient) save(name string, data io.Reader, options ...storage.Optio
...
 
-	const align int64 = 256 << 10 // 256 KiB (both min size and alignment)
+	ctx := context.Background()
+	w := g.bucketHandle.Object(path.Join(g.cfg.Prefix, name)).NewWriter(ctx)
+	if g.cfg.parallelUploadEnabled() {
+		pu := g.cfg.ParallelUpload
+		if g.log != nil && opts.UseLogger {
+			partSize := "SDK default"
+			if pu.PartSize > 0 {
+				partSize = fmt.Sprintf("%v (%v)", pu.PartSize, storage.PrettySize(int64(pu.PartSize)))
```
boris-ilijic: `storage.ComputePartSize` should be used for grps also

pbm/restore/physical.go
```diff
@@ -120,6 +120,7 @@ type PhysRestore struct {
 
 	stopHB        chan struct{}
 	stopCleanupHB chan struct{}
+	hbWG          sync.WaitGroup
```
boris-ilijic: Rather create local storage instance just for the HB and close it when HB is stopped. So, that we do not need to do this type of synchronization.

pbm/restore/logical.go
```diff
@@ -87,6 +87,22 @@ type oplogRange struct {
 	storage storage.Storage
 }
 
+type storageReadCloser struct {
```
boris-ilijic, line 90: It seems that it necessary to refactor logical restore to use single storage instance. This is workaround that's hard to follow, and I am not sure that it closes all flows.

jcechace: That would be basically reverting https://github.com/percona/percona-backup-mongodb/pull/703/changes

I think the reason why each file get's it's own storage is shared state in Downloaders. We would be sharing this field `arenas   []*storage.Arena`

boris-ilijic (dismissed):

boris-ilijic (approved):

PR #1347, opened 2026-07-19
PBM-1700 Limit logical backup stats concurrency

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1700

### DIFF
--- a/pbm/backup/logical.go
+++ b/pbm/backup/logical.go
@@ -27,6 +27,8 @@ import (
 	"github.com/percona/percona-backup-mongodb/pbm/version"
 )
 
+const namespaceStatsConcurrency = 50
+
 func (b *Backup) doLogical(
 	ctx context.Context,
 	bcp *ctrl.BackupCmd,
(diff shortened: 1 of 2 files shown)


pbm/backup/logical.go
```diff
@@ -443,24 +445,24 @@ func getNamespacesSize(ctx context.Context, m *mongo.Client, nss []string) (map[
 	}
 
 	isSelected := util.MakeSelectedPred(nss)
-
 	mu := sync.Mutex{}
-	eg, ctx := errgroup.WithContext(ctx)
+	colEg, colCtx := errgroup.WithContext(ctx)
+	colEg.SetLimit(concurrency)
+	dbEg, dbCtx := errgroup.WithContext(ctx)
+	dbEg.SetLimit(concurrency)
```
boris-ilijic: I'd go here with bigger numbers. For `numParallelCollections` parameter PBM does actual work (backup particular collection), but here PBM just waiting the server which does actual work. 

Imo, we should hardcode something between 50-100 or use some multiplication factor. Benchmark should give us guidelines.

boris-ilijic (approved):

PR #1349, opened 2026-07-21
PBM-1780 remove gcs hmac client

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1780


pbm/storage/gcs/google_client.go
```diff
@@ -4,83 +4,34 @@ import (
 	"context"
```
boris-ilijic: We don't need `google_client.go` file anymore, everything should be placed in `gcs.go`.

boris-ilijic (dismissed):

boris-ilijic (approved):

PR #1293, opened 2026-03-27
PBM-1731: Add backup lifecycle management with GFS rotation to pbm

daniel-percona:
## Summary
This PR introduces **Backup Lifecycle Management** for Percona Backup for MongoDB (PBM). This feature automates the retention and rotation of snapshots using a Grandfather-Father-Son (GFS) policy, addressing critical needs for cost-efficient storage management while ensuring long-term compliance and data durability.

## Key Features
* **Dual Rotation Strategies:**
    * `Rolling` (Default): A time-decay algorithm that creates age-based buckets, ensuring resilience against skipped backup windows.
    * `Calendar`: A strict anchoring strategy for compliance-heavy environments (e.g., "Keep every Friday backup").
* **Safety Circuit Breaker (`minKeep`):** Prevents the "Empty Storage Paradox." If a rotation would leave the cluster with fewer than `X` backups (default 1), the operation aborts to protect the last remaining recovery points.
* **Intelligent Failed Backup Handling:** Optional purging of `StatusError` backups after they age out of the daily window, while protecting in-progress backups.
* **Interactive Safeguards:** Built-in `--dry-run` capability and a mandatory confirmation prompt (default `true`) to prevent accidental data loss during manual runs.
* **Chronological Purging:** Deletions are executed from **oldest to newest**, ensuring the most recent data is preserved longest in the event of a network or process interruption.

## Technical Implementation Details
* **PITR Safety:** The lifecycle manager utilizes the core `backup.DeleteBackup` function, inheriting PBM's internal `isRequiredForOplogSlicing` checks. Deletion will be refused if a snapshot is required for an active Point-in-Time Recovery chain.
* **Concurrency Safe:** The manager strictly filters for `done` snapshots, ensuring that `starting`, `running`, or `stuck` backups are never targeted for deletion.
* **Standardized URI Handling:** Fully compatible with existing `PBM_MONGODB_URI` environment variables and standard PBM connection string formats.

## Testing Performed
* Validated GFS bucket logic across 2 months of synthetic backup history.
* Verified `minKeep` abort logic in both interactive and automated (`prompt: false`) modes.
* Confirmed `Ctrl+C` signal handling during the deletion loop to ensure graceful termination.
* Dry-run verification of "Calendar" vs "Rolling" selection logic.

---

### Configuration Parameters
| Option | Type | Default | Description |
| :--- | :--- | :--- | :--- |
| `lifecycle.enabled` | Boolean | `false` | Toggles automated lifecycle management. |
[...]

PR #1359, opened 2026-08-23
PBM-1731: lifecycle management

jcechace:
**Ticket**: https://perconadev.atlassian.net/browse/PBM-1731

**includes/follows up on**: https://github.com/percona/percona-backup-mongodb/pull/1293

Detailed feature description: https://gist.github.com/jcechace/3706e5eb4ef236e30e4cf35d1ad8d3cc


pbm/config/config.go
```diff
@@ -69,6 +69,90 @@ func validateConfigKey(k string) bool {
...
+		return errors.New("lifecycle.weeklyDay must be between 0 and 6")
+	}
+	if strategy == LifecycleStrategyCalendar && c.MonthlyRetention > 0 &&
+		(c.MonthlyDay < 1 || c.MonthlyDay > 31) {
+		return errors.New("lifecycle.monthlyDay must be between 1 and 31")
+	}
+
+	return nil
+}
+
```
boris-ilijic: I'd move this block to the end of the file to keep `type Config struct {...}` at the top.
It's possible to extract it into separate one, like it was done within `ixquorim.go`

pbm/config/config.go
```diff
@@ -79,7 +163,8 @@ type Config struct {
 	Backup  *BackupConf  `bson:"backup,omitempty" json:"backup,omitempty" yaml:"backup,omitempty"`
 	Restore *RestoreConf `bson:"restore,omitempty" json:"restore,omitempty" yaml:"restore,omitempty"`
 
-	Epoch bson.Timestamp `bson:"epoch" json:"-" yaml:"-"`
+	Epoch     bson.Timestamp `bson:"epoch" json:"-" yaml:"-"`
+	Lifecycle *LifecycleConf `bson:"lifecycle,omitempty" json:"lifecycle,omitempty" yaml:"lifecycle,omitempty"`
```
boris-ilijic, line 81: Should be in the upper section, below `Restore`.

pbm/config/config.go
```diff
@@ -729,6 +831,34 @@ func SetConfigVar(ctx context.Context, m connect.Client, key, val string) error
 	return errors.Wrap(err, "write to db")
 }
 
+func isLifecycleKey(key string) bool {
+	return strings.HasPrefix(key, "lifecycle.")
+}
+
+func applyLifecycleConfigVar(cfg *LifecycleConf, key string, value any) {
```
boris-ilijic, line 754: With this validation approach, the generic purpose of `SetConfigVar` goes in a different direction, opposite to the original idea (and that was the case even before this change, in a smaller scope). I suggest creating a task to rethink and refactor it.

cmd/pbm/delete.go
```diff
@@ -322,10 +357,105 @@ func doCleanup(ctx context.Context, conn connect.Client, pbm *sdk.Client, d *cle
...
+	if err := d.profile.Validate(ctx, conn); err != nil {
+		return nil, err
+	}
+	if !d.dryRun {
+		if err := checkForAnotherOperation(ctx, pbm); err != nil {
+			return nil, err
+		}
+	}
+
+	lifecycleAt := bson.Timestamp{T: uint32(time.Now().UTC().Unix())}
```
boris-ilijic:
```suggestion
	lifecycleAt := bson.Timestamp{T: uint32(time.Now().Unix())}
```

cmd/pbm-agent/delete.go
```diff
@@ -294,6 +323,41 @@ func (a *Agent) Cleanup(ctx context.Context, d *ctrl.CleanupCmd, opid ctrl.OPID,
 	}
 }
 
+func (a *Agent) cleanupLifecycle(
+	ctx context.Context,
+	d *ctrl.CleanupCmd,
+	node string,
+	l log.LogEvent,
+) {
+	lifecycleAt := time.Unix(int64(d.LifecycleAt.T), 0).UTC()
```
boris-ilijic, line 332: `Unix()` is already UTC.

pbm/lifecycle/manager.go
```diff
@@ -0,0 +1,854 @@
...
+	previousRestoreTime bson.Timestamp
+	restoreTime         bson.Timestamp
+}
+
+func (r *Report) addKeepReason(name, reason string) {
+	for _, existing := range r.KeepReasons[name] {
+		if existing == reason {
+			return
+		}
+	}
```
boris-ilijic:
```suggestion
	if slices.Contains(r.KeepReasons[name], reason) {
		return
	}
```

pbm/lifecycle/manager.go
```diff
@@ -0,0 +1,854 @@
...
+	for _, b := range r.BackupsPurged {
+		bType := r.BackupTypes[b] // Fetch the type
+		res += fmt.Sprintf("  - %s <%s>\n", b, bType)
+	}
+	return res
+}
+
+// Evaluate analyzes backups according to the lifecycle configuration and
+// returns a report describing which backups should be kept or purged.
+func Evaluate(cfg config.LifecycleConf, backups []backup.BackupMeta, now time.Time) *Report {
```
boris-ilijic: The function is not used, it is referenced only in tests, so let's move it there if we need it.

pbm/lifecycle/manager.go
```diff
@@ -0,0 +1,854 @@
...
+	weeklyCandidates := make(map[string][]backup.BackupMeta)
+	monthlyCandidates := make(map[string][]backup.BackupMeta)
+
+	for _, bcp := range selectedBackups {
+		if bcp.Status.IsRunning() {
+			continue
+		}
+
+		bcpTime := time.Unix(bcp.StartTS, 0).UTC()
+		if util.IsSelective(bcp.Namespaces) {
```
boris-ilijic, line 627: Lifecycle rules are not applied to selective backups.

I can see in the docs that this is by design, and I understand the technical motivation, but I can imagine a use case where only selective backups are taken. In that case, a config parameter could help (easiest approach).

In general, this should go in limitations, and we can create a ticket later.

pbm/lifecycle/manager.go
```diff
@@ -0,0 +1,854 @@
...
+	if distance < 0 {
+		distance = -distance
+	}
+	return int(distance / (24 * time.Hour))
+}
+
+func betterCandidate(candidate, current *backup.BackupMeta) bool {
+	return current == nil ||
+		candidate.StartTS > current.StartTS ||
+		candidate.StartTS == current.StartTS && candidate.Name < current.Name
```
boris-ilijic: It's better to use `LastWriteTS` here, that means restore time, and that is point in time that represents backup.

boris-ilijic: If LastWriteTS is equal (unlikely but ok), as the second criteria I'd select one with more recent `StartTS` as the better candidate to keep.

pbm/lifecycle/manager.go
```diff
@@ -0,0 +1,844 @@
...
+	}
+
+	report.applyMinKeepGuard(selectedBackups)
+	report.sortDeleteTargets(selectedBackups)
+	return report, nil
+}
+
+// buildIncrementalChains groups incremental metadata by its resolved base and
+// marks chains unsafe when they cross profiles or data-owning storage boundaries.
+func buildIncrementalChains(
```
boris-ilijic: Not sure why we are dealing here with `allBackups`, we should target just backups on the specific storage. 

All incremental chain is stored on the [single storage](https://docs.percona.com/percona-backup-mongodb/features/multi-storage.html?h=profile#delete-backups:~:text=The%20whole%20incremental%20chain%20must%20be%20stored%20on%20the%20same%20storage.%20To%20change%20the%20storage%20for%20incremental%20backups%2C%20you%20must%20start%20a%20new%20backup%20chain%20with%20the%20incremental%20base%20backup%20on%20the%20new%20storage.).

pbm/lifecycle/manager.go
```diff
@@ -0,0 +1,844 @@
...
+func isMainPITRBaseSnapshot(bcp backup.BackupMeta) bool {
+	return !bcp.Store.IsProfile &&
+		bcp.Status == defs.StatusDone &&
+		bcp.Type != defs.ExternalBackup &&
+		!util.IsSelective(bcp.Namespaces)
+}
+
+// applyPITRDeleteChecks ensures lifecycle never proposes a deletion that the
+// checked delete-backup path would reject for PITR safety.
+func (r *Report) applyPITRDeleteChecks(
```
boris-ilijic, line 257: I would take different approach:
- delete all backups that fulfill lifecycle criteria.
- delete all pitr chunks that are hanging without base backup.

boris-ilijic (approved):

PR #1366, opened 2026-08-28
PBM-1776 Fix physical restore log buffer storage lifetime

jcechace:
Ticket: https://perconadev.atlassian.net/browse/PBM-1776

Addresses the regression described in comments

### DIFF
--- a/pbm/restore/physical.go
+++ b/pbm/restore/physical.go
@@ -1086,9 +1086,9 @@ func (n nodeStatus) isFailed() bool {
 // and reset buffer when logs size hist the limit.
 type logBuff struct {
 	buf   *bytes.Buffer
+	stg   storage.Storage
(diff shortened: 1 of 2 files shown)


pbm/restore/physical.go
```diff
@@ -1121,32 +1121,43 @@ func (l *logBuff) flush() error {
 
 func (l *logBuff) Flush() error {
```
boris-ilijic, line 1122: FlushAndClose()

boris-ilijic (approved):

pbm/restore/physical.go
jcechace: I would keep Flush or rename to Close in order to keep standard interface methods. 

Merging, we can adjust later.

